PR #50, opened 2016-09-20
Direct selector

joaomilho:
- Adds the Selector.Direct
- Renames Selector to Selector.Options
- Renames Arrow to Chevron and adds Chevron.Down/Up/Left/Right (naming based on bootstrap icons)


icons/example.jsx
```diff
@@ -87,7 +89,7 @@ export default {
     )
   }, {
     title: 'Small',
-    require: icons.tiny.map(({ name }) => `import ${name} from '@klarna/ui/icons/${name}'`).join('\n'),
+    require: icons.tiny.map(({ displayName }) => `import ${displayName} from '@klarna/ui/icons/${displayName}'`).join('\n'),
```
xaviervia: I'm seeing the output of this as `import undefined from '@klarna/ui/icons/undefined` for every icon. I think since we are not setting it manually, it's just `name`.

icons/Chevron.jsx
```diff
@@ -15,18 +15,34 @@ export default function Arrow ({ color, styles, className, ...props }) {
...
+Chevron.propTypes = {
   color: PropTypes.oneOf(colors),
   styles: PropTypes.object
 }
+
+// The geometrical center would be starting at x = 8, y = 6, but because of
+// https://medium.com/@erqiudao/the-play-button-is-not-optical-alignment-4cea11bda175#.3p6p4iry8
+// it should be 1px off
+
+export const Right = (props) =>
```
xaviervia: It would be nice to add display names for this with a more fully qualified name. Currently the output in the showroom is `<Up />`, `<Down />` and so on. I'd prefer having `<Chevron.Up />` and so on.

icons/example.jsx
joaomilho: ?

icons/Chevron.jsx
joaomilho: yup!

icons/example.jsx
xaviervia: screen shot 2016-09-20 at 20 12 48

joaomilho: I see

PR #54, opened 2016-09-21
Add templates spike

xaviervia:
The idea is to incorporate consolidated templates following the common UIs. Some challenges:

- We need to create a structure for showcasing them easily in the showroom.
- We need to add the necessary missing components, such as Button.Tertiary
- We need to decide whether the Dialog structure should be part of the template or not, and how to do animations in the transition between different dialog contents (with their different sizes)


xaviervia:
screen shot 2016-09-21 at 17 34 11 2

xaviervia:
Note: there is currently no way to open or close the Dialog template right now. Just navigate away. @hannah-ahl will reestructure how we present these templates so that they can be seen both inlined and as a dialog.

xaviervia:
So, I simplified the integration by adding a `<Wrapper>` component for templates that is aimed at keeping how the templates are inlined as an abstract solution. The idea is that if we in the future decide that dialog experiences should be rendered in a complete different manner (say, as a menu floating from the side instead of a FSO) we can do that without changing everyone's integration, just be updating the archetype implementation.

Wdyt?

screen shot 2016-09-27 at 16 54 10

xaviervia:
OK, regarding the previous comment, `Wrapper` will aim at deducing the archetype from the template used inside. So, there will be no `archetype` prop to start with.

xaviervia:
Several features are added to the `Landing`:

screen shot 2016-09-28 at 14 32 19
screen shot 2016-09-28 at 14 32 25
screen shot 2016-09-28 at 14 32 31

xaviervia:
Explanation template:

screen shot 2016-09-28 at 14 48 30

xaviervia:
Selection template: 

screen shot 2016-09-28 at 15 10 05

PR #61, opened 2016-09-23
Fix List.Unordered bullet styles

xaviervia:
Initial option:

screen shot 2016-09-23 at 15 21 09

@hannah-ahl

### DIFF
--- a/List/Item/index.jsx
+++ b/List/Item/index.jsx
@@ -2,13 +2,18 @@ import React from 'react'
 import classNamesBind from 'classnames/bind'
 import defaultStyles from '../styles.scss'
 
-const baseClass = 'list--item'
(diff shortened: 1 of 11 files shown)


xaviervia:
@hannah-ahl we updated and now the styles looked like we agreed before. There is only one caveat though: for the nested lists, we didn't implement the default numbering since it's not extremely straightforward. I opened a discussion for that in the issue https://github.com/klarna/ui/issues/63 since we could have other alternatives.

This is how it looks like now. 

screen shot 2016-09-26 at 11 31 49

hannah-ahl:
Bullets look great!
Hannah Ahl
Product Designer
Product
Klarna AB
Reg no: 556737-0431

Issue #45, opened 2016-09-06
Consider setting the default ones as Uncontrolled


xaviervia:
So, the idea is that simplify the consumption. In that sense, all the components could be passed through the enhancer HOCs (such as stateful focus, stateful value, themeable) and getting the controlled version would be done like in default in React:

- Setting a `defaultValue` will allow to set a value that the user can modify at will, but setting a `value` prop will mean that the component value is now Controlled.
- Setting a `autoFocus` will allow the focus to be moved out, but setting a `focus` prop will mean that the component focus is now Controlled.
- `Theme` should not get in anyones way if not used.

xaviervia:
As an update: the `uncontrolled` decorator added recently is the first drop in this direction. I hope to follow up soon in this issue.

xaviervia:
Continues in #121

PR #77, opened 2016-10-03
(WIP) Make auto controlled

xaviervia:
This is a PoC of implementing https://github.com/klarna/ui/issues/45

Apparently the Controlled vs Uncontrolled as entry point level APIs are creating a lot of issues for newcomers to the library. It could make sense to provide Uncontrolled by default and switch to Controlled if a specific prop is passed, such as React does with `<input>`.

### DIFF
--- a/Field/example.jsx
+++ b/Field/example.jsx
@@ -9,7 +9,7 @@ const Single = {
 import UncontrolledField from '@klarna/ui/uncontrolled/Field'`,
 
   Regular: (
-    <Field label='Enter your email' />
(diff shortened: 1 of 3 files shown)


xaviervia:
OK, some comments in this approach:

1. Use full variable names
2. Why do we expect the APIs from the Selectors to look different? Shouldn't the event be always the native event? If so, what about focus? Could it be native unless we pass a "selector" function?

xaviervia:
Was partially implemented in https://github.com/klarna/ui/pull/89, will keep track of progress in https://github.com/klarna/ui/issues/45

PR #85, opened 2016-10-06
Add second wave of templates

xaviervia:
Ideally will also include an improvement in the way the templates are displayed in the Showroom, following the designs from @hannah-ahl 

To do:

- [x] Fix the `Wrapper` component (done in #97)
- [x] _Make the `Button.Secondary` in `Centered` be a new Tertiary button._ (will be done in https://github.com/klarna/ui/issues/105)
- [x] Fix the spacing in the Input in the `FilteredSelection`
- [x] _Center the Input in `FilteredSelection`_ (actually it was left aligned)
- [x] _Fix the autohiding of the looking glass icon in the Input of `FilteredSelection`_ (won't be done, just kept in place)
- [x] Generalize the ThreeStepExplanation to be XStepExplanation
- [x] _Come up with a Form API and use it in the more complex form-like templates_ (will actually be done on a separate PR after the `<Form>` component is implemented)
- [x] _Use the _themeable_ version for all the components within templates_ (this will be so automatically once we merge #98
- [x] _In the XStepExplanation, center either the text or the icon vertically if the sizes don't match._ (will be done in https://github.com/klarna/ui/issues/107)
- [x] _Contemplate adding logos in the Selection API_ (will be done in https://github.com/klarna/ui/issues/104)


xaviervia:
Added Iconic List component that is used in one of the Templates:

screen shot 2016-10-06 at 13 46 34

xaviervia:
Two more templates added: 

**ConfirmData**:

screen shot 2016-10-06 at 17 52 17

**ThreeStepExplanation**

screen shot 2016-10-06 at 17 54 09

xaviervia:
Add SingleInputPrompt template. This template incorporates a new API for uncontrolled components:

screen shot 2016-10-07 at 16 29 04

xaviervia:
Added small Search icon:

screen shot 2016-10-10 at 11 02 02

xaviervia:
Add Search icon for Inputs and Fields:

screen shot 2016-10-10 at 11 03 18
screen shot 2016-10-10 at 11 03 29

xaviervia:
Added first drop of FilteredSelection template. API complete.

xaviervia:
For the templates with several fields we need a sensible way of doing "Form Component" APIs, since the template will in fact be a form. This connects with the idea of having a Form component available that takes field metadata and constructs forms inside: the API should follow the same pattern.

And idea is to have the `onChange` event called always with the first argument being the key of the field that is being modified, like: 

`('firstname', 'Jane')`

xaviervia:
As a follow up for the Button styles, there is this issue: https://github.com/klarna/ui/issues/91

xaviervia:
FilteredSelection is here:

screen shot 2016-10-11 at 17 13 28

While it is controlled or uncontrolled depending on usage, the list filtering is done by the consumer since we might want to do it with a backend call or so on.

xaviervia:
CenteredSelection template. Only two to go for this iteration.

screen shot 2016-10-11 at 17 37 36

xaviervia:
ReviewData template:

screen shot 2016-10-11 at 19 51 21

One to go.

List/example.jsx
```diff
@@ -97,6 +97,60 @@ export default {
...
+            </List.Iconic.Item>
+            <List.Iconic.Item
+              icon={<svg
+                width='60px'
+                height='60px'
+                viewBox='0 0 60 60'>
+                <polygon
+                  stroke='#979797'
+                  fill='#D8D8D8'
+                  points='30 46.5 12.95 55.46 16.20 36.48 2.41 23.03 21.47 20.26 30 3 38.52 20.26 57.58 23.03 43.79 36.48 47.04 55.46'
```
joaomilho: Since this is just an example, I would favor simpler SVGs, like a circle with a line... something simple 😄

templates/XStepExplanation/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+const baseClass = 'three-step-explanation'
+
+const classes = {
+  list: `${baseClass}__list`
+}
+
+const classNames = classNamesBind.bind(defaultStyles)
+
+export default function XStepExplanation ({
```
joaomilho: We have a help thing that looks very different (and better, I would say) on KDS. Maybe the designers are not in sync about that, I donno. 

screen shot 2016-10-13 at 5 34 48 pm
screen shot 2016-10-13 at 5 34 40 pm
screen shot 2016-10-13 at 5 34 32 pm

templates/examples/Prompts.jsx
```diff
@@ -0,0 +1,72 @@
...
+    //         accept='Continue'
+    //         legal='Chia williamsburg subway tile vaporware, live-edge kinfolk cardigan prism deep v retro seitan.'
+    //         label='Name'
+    //         onChange={(e) => console.log(e.target.value)}
+    //         onAccept={() => console.log('accept')}
+    //         value='Penelope Clearwater'
+    //       />
+    //     }
+    //   }
+    // }
```
joaomilho: ?

templates/examples/Selection.jsx
```diff
@@ -0,0 +1,124 @@
...
+            summary='Photo booth distillery man bun, bitters stumptown freegan cliche cronut green juice.'
+            label='Start typing a flavor'
+            options={[
+              {key: 'chocolate', label: 'Chocolate'},
+              {key: 'vanilla', label: 'Vanilla'},
+              {key: 'raspberry', label: 'Raspberry'},
+              {key: 'tiramisu', label: 'Tiramisu'},
+              {key: 'mascarpone', label: 'Mascarpone'},
+              {key: 'banana-split', label: 'Banana Split'},
+              {key: 'anana', label: 'Anana'}
```
joaomilho: I wonder why we need to do these key/label objects... I guess it's because JS doesn't have proper pairs. I'm planning to add 'em to ion BTW

templates/XStepExplanation/index.jsx
joaomilho: @zenona @yatirk

xaviervia: 👍 

The thing is that we don't support onSwipe anywhere currently and I didn't want to commit to having that with the fix deadline. The Sketch has an example that is closer to the one you show with several steps (although white background) but we scoped it out for the time being.

Worst case scenario, we can do this next drop.

templates/examples/Selection.jsx
xaviervia: The real reason is ordering though, since it's not guaranteed in object structures :D

List/example.jsx
xaviervia: But it's so pretty! 🌟

PR #89, opened 2016-10-07
Add HOC to make anything Uncontrolled on demand

xaviervia:
This is **Part 1** of the saga of moving towards [controlled vs uncontrolled](https://github.com/klarna/ui/issues/45) being a consequence of the provided props instead of a feature of the component. This HOC allows for converting a fully controlled set of props in component into pseudo controlled. The typical case would look like this:

```js
import Field from '@klarna/ui/Field'
import uncontrolled from '@klarna/ui/decorators/uncontrolled'

export default uncontrolled(Field, {
  prop: 'value',
  defaultProp: 'defaulValue',
  handler: 'onChange'
})
```

It will also work for making focus pseudo controlled like:

```js
import Field from '@klarna/ui/Field'

export default uncontrolled(Field, {
  prop: 'focus',
  defaultProp: 'autoFocus',
  handler: 'onFocus',
  reset: 'onBlur'
})
```

Merging this is completely safe since it's not yet used, but I'm planning on setting this as the default for Templates so that consumption of those is easier and it would be nice to get feedback on what you think of this approach.

### DIFF
--- a/decorators/uncontrolled.jsx
+++ b/decorators/uncontrolled.jsx
@@ -1,0 +1,62 @@
+import React, { Component } from 'react'
+
+export default function uncontrolled (Component, {defaultProp, prop, handler, reset}) {
+  return class Uncontrolled extends Component {
+    componentDidMount () {
+      this.setState({
+        [prop]: this.props[prop] != null
+          ? this.props[prop] || ''
+          : this.props[defaultProp]
+      })
+    }
+
+    handleHandler (handler, e) {
+      if (this.props[prop] != null) {
+        handler && handler(e)
+      } else {
+        this.setState({
+          [prop]: e && e.target
+            ? e.target.value
+            : e
+        })
+
+        handler && handler(e)
+      }
+    }
+
+    handleReset (handler, e) {
+      if (this.props[prop] != null) {
+        handler && handler(e)
+      } else {
+        this.setState({
+          [prop]: undefined
+        })
+
+        handler && handler(e)
+      }
+    }
+
+    render () {
+      const consumerHandler = this.props[handler]
+      const consumerReset = this.props[reset]
+
+      const props = Object.keys(this.props)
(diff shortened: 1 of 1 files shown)


decorators/uncontrolled.jsx
```diff
@@ -0,0 +1,62 @@
...
+        handler && handler(e)
+      }
+    }
+
+    render () {
+      const consumerHandler = this.props[handler]
+      const consumerReset = this.props[reset]
+
+      const props = Object.keys(this.props)
+        .filter((key) => key !== handler)
```
cskeppstedt: Should also be filtering the 'reset' prop

decorators/uncontrolled.jsx
```diff
@@ -0,0 +1,62 @@
...
+        .reduce((filteredProps, key) => ({
+          ...filteredProps,
+          [key]: this.props[key]
+        }), {})
+
+      const handlerProps = {
```
cskeppstedt: As you said. You can get rid of both the .filter and this separate handler props mapping by doing the work inside the reduce.

decorators/uncontrolled.jsx
```diff
@@ -0,0 +1,62 @@
...
+    componentDidMount () {
+      this.setState({
+        [prop]: this.props[prop] != null
+          ? this.props[prop] || ''
+          : this.props[defaultProp]
+      })
+    }
+
+    handleHandler (handler, e) {
+      if (this.props[prop] != null) {
```
cskeppstedt: Logic can be simpleified. Invert the `if`, put the handler-stuff always in the bottom

cskeppstedt:
Looks awesome, 👍

xaviervia:
@cskeppstedt Thanks for the review :D Fixed all the issues in af11f8e and 66e90a9 . I will merge as soon as it's passing and then start using it in the Templates to see how it works in there.

PR #90, opened 2016-10-07
Style input placeholder

Deschtex:
In order for us to start using the native placeholder as a compliment to the label, it needs to be styled to look like the label.

### DIFF
--- a/Field/styles.scss
+++ b/Field/styles.scss
@@ -183,6 +183,18 @@
     @include typography(map-get($font-sizes, big-body-desktop), semi-bold);
   }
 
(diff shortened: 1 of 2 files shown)


Field/styles.scss
```diff
@@ -183,6 +183,18 @@
     @include typography(map-get($font-sizes, big-body-desktop), semi-bold);
   }
 
+  &::-webkit-input-placeholder {
```
Deschtex: It's unfortunately not possible to chain this selector with `:-ms-input-placeholder`. That's why they're identical.

Issue #64, opened 2016-09-26
Incorporate better wide dialogs


xaviervia:
See https://github.com/klarna/ui-css-components/pull/69

xaviervia:
So, notes:

- [ ] In mobile landscape, the dialog should wrap to the sides, and the padding should be bigger. How big is TBD.
- [ ] We keep the icons as usual in the upper right corner 30px from top and 30px from right.
- [ ] We assume that text will be bigger also in landscape for dialogs.
- [ ] Test the extremes of the landscape: smallest landscape phone vs iPhone 6 Plus. What exactly is the breakpoint (in pixels)? It should take into consideration both width and height.
- [ ] We assume that Tablet will need a different solution.
- [ ] We need a breakpoint for a desktop with extreme width and little height. For that scenario we should probably have the dialog as usual but without rounded corners (@jonotrujillo 's last implementation in the linked PR)
- [ ] Add these sizes in Sketch so that we can try out the designs before implementing.

- In mobile, we want to always cover the screen

Issue #20, opened 2016-08-23
Wrapper needs keyboard interaction


xaviervia:

Nevon:
What does this mean? You mean tabindexes for the close icon etc? Something I'm missing is to be able to close it by pressing escape, but since it's a fully uncontrolled component, I'm not sure how we would model that - or if it should be left up to the applications to implement themselves.

xaviervia:
Both.

This as a placeholder issue, we can outline the requirements in more detail. What comes to mind is:

- Close and Back icons in tabindex
- Esc key to close the dialog
- Clicking in the overlay closes the dialog

In a way, actually closing the dialog would be (for now) left to the applications.... although I guess we could make this a pseudo controlled component with the uncontrolled decorator. what we need to add is the ability to _capture_ the esc key and clicks in the overlay and all that.

xaviervia:
I wonder how could this even be done. For key board interactions to be captured in the context of the Dialog, the events have to listened to in the `window` or some element within the Dialog needs to be in focus. Either way, and in the absence of some hack, it doesn’t feel that the Dialog itself can capture these events. Maybe this is something that unfortunately consumers will have to do on their own listening to `window.addEventListener('keydown')` and using the information to close the Dialog when it’s open.

Nevon:
Why couldn't the dialog listen on the window?

xaviervia:
Isn't that rather weird? Separation of concerns and all that.

Nevon:
I don't think you mean separation of concerns, but I think I understand what you mean. It's listening in on global events, but I think it's okay. If you press escape when a modal is open, we want to close it regardless of where your focus was at the time.

xaviervia:
I like this line of reasoning. I guess an escape key pressed on while the input is on a field could be intercepted by the the field with stopPropagation so that the window doesn't capture it and the Dialog is closed.

xaviervia:
BTW this is a feature of the `Wrapper`, not the "Dialog". For starters, the Dialog does not (and never did) provide an `onClose` handler, while the Wrapper does. Also the Dialog itself is just a layer inside the full screen context, which lives alongside the Backdrop and potentially other layers. I'll rename the ticket.

And in this case, what you say makes even more sense. The only annoyance we will have is that pressing "escape" when the template demos are open will fire the `onClose` function of all of them.

While we are at it, should we provide a key shortcut for the dialog "back" button? It feel that that should be probably tied to the browser navigation.

[...]

PR #94, opened 2016-10-10
Rebuild dialog implementation

xaviervia:
Following @Deschtex specifications.


xaviervia:
https://github.com/klarna/ui/issues/64 will be followed up in here as well.

xaviervia:
- https://github.com/klarna/ui/issues/20
- https://github.com/klarna/ui/issues/64
- https://github.com/klarna/ui/issues/162
- https://github.com/klarna/ui/issues/192

xaviervia:
Implementation wise, this is looking good, but we found an issue where the interface in iOS jumps when going fullscreen. If that is deemed acceptable (cc @hannah-ahl @jordiparra ) I’ll move forward with this implementation.

xaviervia:
OK, the "jump" was deemed acceptable. I will proceed with this.

xaviervia:
Gathering my thoughts here: the Backdrop and the pinToTop functions could probably an agnostic component in another library, since their behavior is not directly related to the Dialog. The Dialog would implement those functions, and consumers would be encouraged to pull the "pinToTop" directly in the context that they need when they need that one in isolation (such in another window frame). This is also because it doesn't make any sense to include the entire UI for just that small library, especially in small apps that are only meant to bootstrap and position iframes.

xaviervia:
Pending in this branch:

- [ ] Add `topRight` and `topLeft` prop support to IconButtons to fill for the removal of the Dialog.Icon
- [ ] Add a `<Footer>` component capable of positioning itself within the layout with flexbox or positioning itself absolute in the bottom.
- [ ] Add the desktop styles to the Dialog component
- [ ] Test the rebuild of the Wrapper and template implementations with this new Dialog
- [ ] Rebuild all Dialogs on top of this new API.

xaviervia:
Dialogs are going to be "inlineable" now, since the dialog itself is just a layer. Combining this with the fact that styling is going to be done through inline styles, it means that we probably want to set up all the "responsive modes" for dialogs as properties to the dialogs themselves. These modes are:

- Narrow mobile (< 320 width): takes over whole container size, no rounded borders, icons get adjusted to a certain distance from the top left and top right.
- Regular mobile (< 569 width): takes over the whole container, no rounded borders, icons get adjusted to a different distance from the top left and top right.
- Regular mobile landscape (arbitrary width, height < 569): same as Regular mobile.
- Desktop: fix width (with `wide` mode), height is `auto`, centered horizontally and vertically in the middle of the screen.

These have to be mapped to props in some way. IconButton's need to be updated as well to support the "small padding" and "big padding" modes.

xaviervia:
Wrapper related tasks:

- [ ] Give it an "visible" and "visible" false state with animations in between
- [ ] Make it autocontrolled with an `autoVisible` flag to show by default and allow `onClose` to be used to close it (will adding the higher-order component cause the close icon to always be rendered :( ? )
- [ ] Provide "onClose" keyboard navigation to the Wrapper

xaviervia:
Template tasks:

- [ ] Add propTypes to template components - https://github.com/klarna/ui/issues/227

xaviervia:
Dialog tasks:

- [ ] Make the default floating mode `wide` and add a `narrow` prop to put it back to the previous size https://github.com/klarna/ui/issues/192

xaviervia:
screen shot 2017-03-03 at 10 24 31

xaviervia:
Lekplats Canvas related tasks:

- [ ] Add landscape mode for all sizes

xaviervia:
This will be done in the next iteration.

PR #95, opened 2016-10-11
Add HoC for MouseflowExclude

Nevon:
I wasn't able to find an example of an exported HoC anywhere, so I just made up a pattern for it. Let me know if you think it should be structured some other way.

### DIFF
--- a/MouseflowExclude/example.jsx
+++ b/MouseflowExclude/example.jsx
@@ -26,7 +26,7 @@ import UncontrolledField from '@klarna/ui/uncontrolled/Field'`,
             name='a-sensitive-valued-field'
             label='Sensitive information'
           />
-        </MouseflowExclude>,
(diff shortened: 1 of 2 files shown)


xaviervia:
Neat. I'd like the usage of the HoC to be closer to the live example though. What do you think of it looking like this?

screen shot 2016-10-11 at 12 55 14

Nevon:
Yeah, that looks great. That's actually what I attempted to do, but I couldn't get the thing to build. Patches welcome!

xaviervia:
There. But I had a trick down the sleeve: you can use a `MANUAL` type example and then pass the code to be displayed. It was added for examples that won't work otherwise.

The only missing thing now would be the highlighting for `const` and so on, but I think we can merge without that since we don't have a color scheme for that (or a format parser for that matter).

Issue #79, opened 2016-10-04
Improvements to Wrapper component


xaviervia:
- [x] Be able to pass className
- [x] Spread props over top level component (allow set an `id`)
- [x] Support `onClose`
- [x] Support `onBack`

xaviervia:
Fixed in #97

PR #97, opened 2016-10-11
Improve wrapper

xaviervia:
Implement the fixes in https://github.com/klarna/ui/issues/79


xaviervia:
OK, so added support for all the features required in the issue. Here is a short example of consumption.

screen shot 2016-10-11 at 14 02 18

Unfortunately I'm not going to be able to show these examples in the Showroom since the current way of displaying them doesn't really allow for fullscreen examples. I left the example live in the http://localhost:7777/ui/#Explanation so you can check it out, but I will delete those commits before merging.

Ping @farfalena

farfalena:
Looks good to me!

PR #98, opened 2016-10-11
Make all components themeable by default

xaviervia:
This is a small API change that will make all the components that support dynamic styling to be themeable by default. _Themeable_ means that, given a `<Theme>` component that set _customizations_ somewhere up in the tree, they will pick up those customizations from the React.context and use them to theme themselves.

A couple of observations:

- The only change visible in the Showroom will be that the Theme demo will require simple components, not the `Themeable` versions
- Uncontrolled components will become themeable as well
- This change is backwards compatible
- But it will also mean that the versions in the `themeable` folder are deprecated and should be removed in v2.


xaviervia:
This is going to be particularly useful to enable theming of components within Templates https://github.com/klarna/ui/pull/85

xaviervia:
As said, the visible difference is that the Theme example in the Showroom now shows that the plain components will work with it directly.

screen shot 2016-10-11 at 15 55 01

Ping @Deschtex wdyt? For your app this will mean that you don't need to use the `themeable` version in consumption, but even if you keep doing, until v2 it should be OK, nothing changed.

Checklist/index.jsx
```diff
@@ -52,31 +59,37 @@ export function Item ({ className, children, customize, styles }) {
...
-Item.displayName = 'Checklist.Item'
+ChecklistItem.displayName = 'Checklist.Item'
 
-Item.propTypes = {
+ChecklistItem.propTypes = {
   className: PropTypes.string,
   children: PropTypes.node,
   styles: PropTypes.object,
   customize: PropTypes.shape({
     strokeColor: PropTypes.string.isRequired
```
JoakimLofgren: No text color here?

Button/Secondary/index.jsx
```diff
@@ -114,3 +115,12 @@ Secondary.propTypes = {
   disabled: PropTypes.bool,
   styles: PropTypes.object
 }
+
+export default themeable(Secondary, (customizations, { customize }) => ({
+  customize: {
+    ...customize,
+    backgroundColor: customizations.color_button,
+    borderRadius: customizations.radius_border,
```
JoakimLofgren: No prop type for this

Theme/example.jsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react'
 import Fieldset from '../Fieldset'
 import Theme from '../Theme'
-import * as Button from '../Button'
-import * as Checklist from '../Checklist'
+import * as Button from '../themeable/Button'
```
JoakimLofgren: You committed the test example for triggering the deprecation notice? :)

xaviervia: Haha indeed. Will remove

xaviervia: Fixed in 0400f8d

Button/Secondary/index.jsx
xaviervia: Fixed in 205e9b6

PR #109, opened 2016-10-13
Preact test

joaomilho:
*** THIS PR IS _NOT_ SUPPOSED TO BE MERGED

So I did it using the shortest path: Using `preact-compat`. It reroutes calls to `react` and `react-dom`, basically. This means, I guess, that people can decide whether to use react or preact in their projects (setting the alias there). We should try in a consumer project.

I had 3 issues:

1 - The `asCode` from `react-element-to-jsx-string` uses `react/lib/TestUtils Something`, but it doesn't work because I'm using a compat layer for preact and it points to `preact-compat/lib/...`. It can be solved by changing this lib to use the new lib for test utils;
2 - The masked input also does something similar;
3 - `replaceState` doesn't seem to work, we use it only in the Showcase index;

It seems to work fine with all other libs, like react-motion, for instance.


xaviervia:
Hm. Maybe we need to homebrew the asCode. The more I work on it the more deficiencies I find with the current approach. After all, it's totally possible to transverse the element structure without using `react/lib/TestUtils`. 

About `replaceState`, that's my bad, I used it even when the documentation explicitly says that it is deprecated. It's simple to replace though, and it should be done anyway.

joaomilho:
I'll scrap this PR and send a new one only changing the README.

joaomilho:
Closed in favour of #139

PR #111, opened 2016-10-13
Extend Fieldset component

xaviervia:
This being first drop, the Form component doesn't have any layout calculations, which would be really nice to have. It is automatically uncontrolled by passing `defaultValues` though.

@JoakimLofgren the API looks slightly different that we talked. Mainly the `fields` are now an Array. I kind of think this makes sense, since the `position` value is a little awkward, and right now _all data_ specified in `fields` is propagated and spread down into the final fields.

screen shot 2016-10-13 at 14 05 50
screen shot 2016-10-13 at 14 05 57
screen shot 2016-10-13 at 14 06 06


Form/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+  onChange,
+  onFocus,
+  values,
+  ...props
+}) {
+  const FieldType = fieldType === 'field'
+    ? Field
+    : Input
+
+  return <Fieldset>
```
joaomilho: 1) I find it weird that `Form` doesn't create a `form`. This is important because of the of the "principle of least surprise". For instance, I may wanna write a test that asserts that I can see a form, and my intuition would be that I can check for a `form` tag;

1.1) I find it odd that there's no onSubmit that gives me a serialized object containing all values. It's very likely one would like to use this component uncontrolled and only get the values on a submit. Actually, the default behaviour should be just like an html form IMO.

Form/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+const handleFocus = (onFocus, name) => () =>
+  onFocus(name)
+
+export default compose(
+  uncontrolled({
+    prop: 'focus',
+    defaultProp: 'autoFocus',
+    handlerName: 'onFocus',
+    handlerSelector: (x) => x,
```
joaomilho: ? can you explain that?

Form/example.jsx
```diff
@@ -0,0 +1,99 @@
...
+          {
+            name: 'address',
+            label: 'Street address'
+          }
+        ]}
+        defaultValues={{
+          firstname: 'Thomas'
+        }}
+        fieldType='input'
+        onChange={(values) => console.log('new values', values)}
```
joaomilho: What exactly onChange receives? Can I know which field changed?

Form/example.jsx
```diff
@@ -0,0 +1,99 @@
...
+            right: true
+          },
+          {
+            name: 'address',
+            label: 'Street address'
+          }
+        ]}
+        defaultValues={{
+          firstname: 'Thomas'
+        }}
```
joaomilho: It seems to me the `defaultValues` shouldn't be an object on its own, but a value on each field object, like:
```js
  name: 'address',
  label: 'Street address',
  defaultValue: 'Dumb street'
```

Form/example.jsx
```diff
@@ -0,0 +1,99 @@
...
+            name: 'firstname',
+            label: 'First Name',
+            size: '1/2',
+            left: true
+          },
+          {
+            name: 'lastname',
+            label: 'Last Name',
+            size: '1/2',
+            right: true
```
joaomilho: It seems redundant to say the 1st one is left and the 2nd is right when I say their size is 1/2 and the order is guaranteed. Can't it be smarter?

Form/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+  fields: PropTypes.arrayOf(PropTypes.shape({
+    name: PropTypes.string.isRequired,
+    label: PropTypes.string.isRequired
+  })),
+  fieldType: PropTypes.oneOf(['field', 'input']),
+  focus: PropTypes.string,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  values: PropTypes.objectOf(PropTypes.string)
```
joaomilho: Same for values, it seem it should be part of the fields.

xaviervia:
Ok, many of the comments address the same underlying issue, that is, the why of the split in the fields vs values API. Well, the idea only makes sense taking into consideration the assumption that we want to split the API into two parts:

- Top down, immutable props sent from the consumer and spread with no modification into the Fields
- "Clever" Form/Fieldset specific configuration that it actually managed by the Form/Fieldset component and it's subject to be affected by user behaviour.

On top of that, the assumption is that the `onChange/value` API signature form a kind of abstract type, as long as the signature of those is that:

```javascript
type InteractiveComponent<T> = {
  onChange: (<T>) => any,
  value: <T>
}
```

This is already true for Field, Input, Switch.Checkbox, all the selector variations (Radio, Selector.Options) long etc. Of course, it can be argued that those components are _atoms_ while the Fieldset/Form is a _molecule_... but treating it as an atom in the API has advantages, among them:

- Consistency. Since all the other components work the same way, this is just a generalization of the same API
- We can reuse the HoC for making a Field uncontrolled. The HoC is trusted and we run into less possible conditions for quirky "uncontrolled" components, which we found several in the past.
- Simplifying documentation and explanation.

There are however some caveats that we tried to address in the discussion with @JoakimLofgren, the main being:

- It actually kind of violates Least Surprise. You would expect to have a single blob of data sent as Fields
- The signature of `onChange` is pretty weird.

The Least Surprise part is arguable: if the mindset is that all interactive components match the type, then the least surprise is that the Form/Fieldset does as well. The signature part is a reasonable trade off in my opinion, but I don't know.

Another reason I personally prefer the `InteractiveComponent` API in this case is that it makes totally clear what props are just going to be spread over the inner components, which gives more clarity to the API, since initially it won't be very smart. Furthermore, with this approach, "smart" improvements can be added on top of the component without being API breaking. A clear example is:

> We want to provide automatic layout. However, in first drop we don't have the time to build that. So we give users the escape hatch of using the manual layout by passing `top`, `left` and all the other layout props in the field objects for the time being.
> When we eventually add support for clever layout however, it would be odd to add it as part of the `fields` structure since that is expected to be one that simply spreads over the inner fields. A more backwards compatible approach would be to add a `layout` prop in the Form/Fieldset itself that takes the structure `{ firstname: NARROW, lastname: NARROW }` and so on. Any extra metadata required to be parsed by the Form/Fieldset can be safely added there, without interfering with consumer expectations.

That's the long explanation of the thinking process, but I'm totally open to refutation of this. Let's discuss the `uncontrolled` after this discussion since that depends entirely of the outcome of this.

Issue #99, opened 2016-10-12
Add version on tag


xaviervia:
The idea is to drive releases by creating Github releases instead of using the `npm version` in the command line. In order to achieve that, we would need to:

1. Ignore (or even better, remove) the `version` entry from the package.json
2. On tag, in Travis, add the version to the package.json that will be sent to npm using the value from the current tag

xaviervia:
Continue in #120

PR #120, opened 2016-10-17
Add a built version

xaviervia:
## Purpose

The MVP for this feature is to make the main package entry point (`import * as UI from '@klarna/ui'`) a built JavaScript file. That is fairly straightforward to support.
## All entry points

The ideal solution however would be to provide built entry points for _each_ of the components, not just the index, but I'm pondering how to do that and keep the webpack configuration manageable. My knee jerk reaction would be to move the project to Sagui, move everthing to the `src` folder and build on `prepublish`, as we do for [most](https://github.com/xaviervia/react-context-props) [other](https://github.com/xaviervia/patch-react-proptypes-add-introspection) JS projects.
## Building on tag

Given that this will imply changing the build and publish system, is probably a great time to introduce [adding version on tag](https://github.com/klarna/ui/issues/99). That is, on travis, we add so locally it updates the package.json to have the version taken from the tag.


JoakimLofgren:
I basically added the following to the webpack config (don't know if everything is needed) when I tried to use the toolkit in ClojureScript.
```json
   library: 'kui',
   libraryTarget: 'umd',
   umdNamedDefine: true
```

I also had to do the following to get it to work with the Google Closure compiler: (in my project)
```bash
npm install -g externs-generator
generate-extern -f index-built.js -n kui -o kui-extern.js
```

project.clj:
```clojure
...
  :cljsbuild {:builds
              [{:id "dev"
...
                :compiler {:main ui-reagent.core
...
                           :externs ["src/kui-extern.js"]
                           :foreign-libs [{:file "src/index-built.js"
                                           :provides ["kui"]}]
...
```

core.cljs:
```ClojureScript
(ns ui-reagent.core
  (:require [reagent.core :as reagent :refer [atom]]
            [kui]))
...

(def primary-button
  (reagent/adapt-react-class (.-Primary kui/Button)))

...

(defn hello-world []
  [:div [:h1 "Hello UI"]
        [primary-button {:size "big"} "Button text"]])

(reagent/render-component [hello-world]
                          (. js/document (getElementById "app")))
```

Displaying a button is not very pretty, but oh well it works with a nice wrapper function :)

JoakimLofgren:
I think you'd need a separate config for the library and the showroom.
So maybe moving the showroom into a folder called "showroom" and having that as a new npm project ?
We did that for our internal components, and basically just added in the package.json: `"@klarna/ui": "file:../"`.
Then you could also build the showroom using the "built" components to make sure that works? :)

Using the `file:/` repository support is pretty shitty, so when developing you'd want to use `npm link`.

But having an entry per component would be great. prepublish and add things into a dist folder?
Like dist/index-built.js, dist/button-built.js etc.

In our package.json we also have:
```
  "main": "./dist/bundle.js",
  "jsnext:main": "src/index.jsx",
```

https://github.com/rollup/rollup/wiki/jsnext:main

xaviervia:
Hm. I'd rather move the project to Sagui then, since we don’t want anything special in the configurations really and I don’t feel like maintaining webpack configurations. It wouldn't even be a breaking change, since the published version would still expose the same structure… but on the other hand it would no longer be possible to pull ui versions from a git branch, that is kind of a downer.

In general though your experiment does mean that exporting in this way we can indeed have consumers that pull the built version, into clojurescript or into whatever else. I guess the MVP would be pretty much what it’s here plus UMD, we can put that configuration behind a flag for the `npm run dist` instead of creating a separate webpack configuration, or?

xaviervia:
@JoakimLofgren actually, it works fine with adding your configuration to the regular `webpack.config.js` . I don't think we need separate configurations. I pushed those changes, wdyt? I’m ready to merge.

PR #121, opened 2016-10-17
Make pseudo controlled the default

xaviervia:
Will solve #45 

## Back to React’s default approach

The idea is to use the same API as React: passing a `value` property will make the components controlled, while leaving `value` undefined or setting a `defaultValue` will make the components uncontrolled.

The same goes for focus: setting the `focus` prop will make the focus of the component controlled, while not setting it or setting the `autoFocus` prop will make the components' focus uncontrolled.

## How the new API look like

For example, say you have the `Field` component and you want it to have a default value and to be automatically focused when page load, but you want the user to be able to type and focus somewhere else without handling the `onChange` and `onBlur` events explicitly. With this change, you can do:

```javascript
<Field defaultValue='Luke' autoFocus />
```

If you instead want to set a value and focus on the Field but want the component to behave as fully controlled, you can do:

```javascript
<Field value='Luke' onChange={...} focus onBlur={...} onFocus={...} />
```

Mind that not specifying these props will mean that the Field is uncontrolled:

```
// You can type and focus in this one:
<Field />

// because it's the same as doing
<Field defaultValue='' autofocus={false} />
```

…so if you want an empty, unfocused fully controlled component, you have to set focus and value to they default values explicitly:

```javascript
// Fully controlled but empty and unfocused
<Field value='' focus={false} />
```

You can also mix and match. A very popular combo is having a fully controlled value but an uncontrolled focus:

```javascript
<Field value='Luke' onChange={...} autoFocus />
```

## Other changes

Taking the opportunity since this changes are API-breaking (although not much for those already using the controlled versions of the components), there are a couple of other improvements bundled in this PR:

- Use the extracted [`@klarna/higher-order-components`](https://github.com/klarna/higher-order-components) for the decorators instead of having them inside.
- Started pulling utility functions such as `compose` from Ramda instead of using local copies of them.
- Remove the `themeable` and `uncontrolled` folders which are now redundant.


xaviervia:
List of Components that are using this type of API (checked are the ones already updated):

- [ ] Dialog (will be addressed in https://github.com/klarna/ui/issues/162)
- [ ] Dropdown
- [ ] Field
- [x] Fieldset
- [ ] Input
- [ ] Installments
- [ ] Radio
- [ ] Selector.Options
- [ ] Switch

(Templates are all updated already)

xaviervia:
- [ ] Fix that some examples still talk about `uncontrolled`

xaviervia:
Show falsy prop in this example:

screen shot 2017-01-27 at 13 46 35

Dropdown/index.jsx
```diff
@@ -249,6 +253,19 @@ const onMouseLeave = (component) => () =>
   })
 
 export default compose(
+  uncontrolled({
+    prop: 'focus',
+    defaultProp: 'autoFocus',
+    handlerName: 'onFocus',
```
cskeppstedt: This is maybe a bit late to the party (especially since I reviewed this implementation once before), but it struck me that maybe the explicit handler/reset duo isn't needed, it could be generalized to this (I _think_):

    handlers: {
      onFocus: () => true,
      onBlur: () => false
    }

Then you'd have the names of the handlers as keys of the `handlers` object, and the handlers readily attached to them. Also you don't need the concept of a resetName + handler, it's all just handlers that affect the prop in some way :) This was just a thought, I think it would work for the current use-cases but haven't gone into depth. Wdyt? I'm not necessarily saying that's a better design or anything, just exploring.

xaviervia: I think your proposal is a clear improvement over what we have today. This generalization feels less _ad hoc_, more future proof and easier to understand.

I will change the higher order components to this approach.

xaviervia:
Ready for another pass if you feel like it @cskeppstedt :)

PR #123, opened 2016-10-17
Fix spinning Loader in Safari

### LINKED ISSUE #92: Loader not spinning in Safari
The Loader it's actually spinning, but the shape renders incorrectly so it doesn't look like spinning.

screen shot 2016-10-10 at 15 15 18

xaviervia:
Fixes #92.

It was surprisingly easy :D -> http://stackoverflow.com/a/26220870/5801727

screen shot 2016-10-17 at 15 13 10


henpe:
👍

filiphi:
👍

PR #130, opened 2016-10-18
Use node 6 in travis for better speed

xaviervia:
screen shot 2016-10-18 at 16 49 16

vs

screen shot 2016-10-18 at 16 49 03

Node 6 =
warp_speed_cat_gif__test_for_larger_project__by_ghostexorcist-d4zmkgg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '4'
+- '6'
 before_script:
 - npm install
 script:


joaomilho:
goddamn

joaomilho:
But how can you prove that the difference is because of Node version, and not from the change from "Fernando Via Canel" to "Xavier Via"?

xaviervia:
philosoraptor

PR #135, opened 2016-10-19
Swipable switch

Nevon:
swipeable-switch

The code isn't the most beautiful thing I've ever written, but it works quite well. Feels A LOT better now than it did before.

The current implementation allows you to start swiping anywhere on the label or the bullet, because the bullet by itself is a rather small target.

Something to consider:

* We had an off-by-one error where you could swipe it 1px "too far", and then it would snap back to the right edge when you let go. I actually felt like this felt really good and tactile, but I'm not sure if that's the direction we want to go with these components.

### DIFF
--- a/Switch/Toggle/index.jsx
+++ b/Switch/Toggle/index.jsx
@@ -9,11 +9,90 @@ const classes = {
   bullet: `${baseClass}__bullet`,
   bulletToggle: `${baseClass}__bullet__toggle`,
   label: `${baseClass}__label`,
-  input: `${baseClass}__input`
+  input: `${baseClass}__input`,
+  container: `${baseClass}__container`
 }
 
-const press = (component) => () => component.setState({ pressed: true })
-const release = (component) => () => component.setState({ pressed: false })
+const pressMouse = (component) => () => component.setState({ pressed: true })
+const releaseMouse = (component) => () => component.setState({ pressed: false })
+const pressTouch = (component) => (e) => {
+  component.setState({
+    pressed: true,
+    touchStartPositionX: e.changedTouches[0].pageX
+  })
+}
+const releaseTouch = (component) => (e) => {
+  const { touchStartPositionX } = component.state
+  const touchPositionX = e.changedTouches[0].pageX
+  const { value } = component.props
+
+  if (touchStartPositionX < touchPositionX && !value) {
+    component.props.onChange && component.props.onChange(true)
+
+  } else if (touchStartPositionX > touchPositionX && value) {
+    component.props.onChange && component.props.onChange(false)
+  }
+
+  if (!component.props.focus) {
+    component.props.onFocus && component.props.onFocus()
+  }
+
+  component.setState({
+    pressed: false,
+    touchPositionX: undefined,
+    touchStartPositionX: undefined,
+    pseudoValue: undefined,
+    bulletPosition: undefined
+  })
+}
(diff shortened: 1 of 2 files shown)


Switch/Toggle/index.jsx
```diff
@@ -86,11 +166,11 @@ const Toggle = React.createClass({
       ...remainingProps
     } = this.props
 
-    const { pressed } = this.state
+    const { pressed, pseudoValue } = this.state
 
     const classNames = classNamesBind.bind({ ...defaultStyles, ...styles })
     const cls = classNames(baseClass, {
-      'is-checked': value,
+      'is-checked': pseudoValue != null ? pseudoValue : value,
```
joaomilho: `!==` ?

Switch/Toggle/styles.scss
```diff
@@ -86,40 +97,11 @@
...
-
-  .right.is-checked & {
-    left: auto;
-    right: ($grid * .2);
-  }
-
-  .right.is-pressed.is-checked & {
-    left: auto;
-    right: ($grid * .2);
+    transform: translateX($grid * 3);
```
joaomilho: IE9?

joaomilho:
awesome

Switch/Toggle/index.jsx
Nevon: The value would be `undefined`, which wouldn't match. `!=` matches both `null` and `undefined`. However, there shouldn't be a scenario where the value is `null`, so we might as well do `!== undefined`.

xaviervia:
@Alexintrino pointed out that we might want to block the scroll event when doing that, since otherwise it could back in the history and so on.

Nevon:
I don't think we actually need to. I've tested it out in both iOS and Android, and the only way it would happen is if you position the switch completely flush against the edge of the screen. In the showroom where it's 20 or so pixels off from the side, I never accidentally tripped the history back function.

xaviervia:
Awesome

xaviervia:
I want to do a quick visual review at this with the designers today since among other things we killed the making-it-wider-on-press. Other than that I'm super happy about it, I think we can iterate on the specific handler code later on when we have the need for other components capturing swipe.

Nevon:
Live demo for the designers to review is available here: https://jsfiddle.net/rso7uh0v/embedded/result/dark/

Open on an actual phone or use touch emulation in the browser for the touch interactions to work.

xaviervia:
Awesome. @hannah-ahl care to review? The change to notice (aside from the sliding) is that the dot in the switch no longers get fat when pressed. Is that OK with you?

Nevon:
So Corbett immediately managed to do the history back gesture as a native iOS user. 😢 The back gesture is implemented in the OS level, so there's no way for us to cancel it (thanks Apple).

The only way we can work around it is to design our interfaces in such a way that the switch is not close to the edge. For situations where the control will be close to the edge, you should use a checkbox.

xaviervia:
Oj, oj, oj, oj

hannah-ahl:
I’d probably ask Jordi or Alexey for this specific case, I find myself too unfamiliar with how the switch worked before to be able to give a judgement
Hannah Ahl
Product Designer
Product
Klarna AB
Reg no: 556737-0431

hannah-ahl:
Gosh darn it
Hannah Ahl
Product Designer
Product
Klarna AB
Reg no: 556737-0431

PR #136, opened 2016-10-19
(WIP) Add calendar icon

JoakimLofgren:
For the upcoming date picker we need a calendar icon.
@hankatt Did we need to update the design any?

Preview of the showroom tiny section:
calendar-icon

### DIFF
--- a/icons/Calendar.jsx
+++ b/icons/Calendar.jsx
@@ -1,0 +1,44 @@
+import React, { PropTypes } from 'react'
+import classNamesBind from 'classnames/bind'
+import defaultStyles from './styles.scss'
+import colors from './constants/colors'
+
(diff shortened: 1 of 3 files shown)


hankatt:
Looking at https://klarna.github.io/ui/#Input, I noticed that this icon is already appearing in the showroom. There should be no need to update the design, assuming it is the same as the one used in the Input example.

JoakimLofgren:
Related issue then https://github.com/klarna/ui/issues/137 :)

xaviervia:
I guess we should reject this one. I think it’s OK to repeat the icons: as we said somewhere else, they are just markup, not "assets".

JoakimLofgren:
Yeah. And I actually wanted to use the Field anyways for the datepicker.

PR #139, opened 2016-10-20
Add preact info

joaomilho:


joaomilho:
That's it

xaviervia:
Should we add this info to the Getting Started page as well? I guess the README and Getting Started should more or less share the same info

xaviervia:
On a more long term comment, we could probably literally put the same information: load the README with a markdown loader, pass it through html2react, :shipit:

joaomilho:
?

joaomilho:
ah, to the showcase, right!

pirelenito:
This change should have been a major.

PR #141, opened 2016-10-24
Add form prompt template

xaviervia:
screen shot 2016-10-24 at 09 57 45

### DIFF
--- a/decorators/uncontrolled.jsx
+++ b/decorators/uncontrolled.jsx
@@ -35,7 +35,13 @@ export default ({defaultProp, prop, handlerName, handlerSelector, resetHandlerNa
 
   render () {
     const props = {
-      ...this.props,
+      ...Object.keys(this.props).reduce((filtered, propName) => {
+        if (propName !== defaultProp) {
+          filtered[propName] = this.props[propName]
+        }
+
+        return filtered
+      }, {}),
       [handlerName]: this.handleHandler.bind(this, this.props[handlerName]),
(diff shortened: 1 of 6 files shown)


decorators/uncontrolled.jsx
```diff
@@ -35,7 +35,13 @@ export default ({defaultProp, prop, handlerName, handlerSelector, resetHandlerNa
 
   render () {
     const props = {
-      ...this.props,
+      ...Object.keys(this.props).reduce((filtered, propName) => {
```
Nevon: Consider breaking this up into an explicit filter step and an explicit copy step to make it clearer.

Nevon: Also, I can't actually comment on it because Github's reviewing functionality sucks, but down there, you're doing a bind within the render function. That's not a good thing, as it causes you to create new functions on every render, which for a component like this could be _a lot_ of times.

templates/examples/Prompts.jsx
```diff
@@ -58,6 +59,55 @@ export default {
...
+    {
+      title: 'CenteredForm',
+      require: 'import CenteredForm from \'@klarna/ui/templates/CenteredForm\'',
+      type: TEMPLATE,
+
+      examples: {
+        Regular: {
+          inline: <CenteredForm
+            accept='Continue'
+            cancel='I rather get a car'
```
Nevon: Should be `I'd rather get a car` 😉

PR #144, opened 2016-10-24
Animate installments

### LINKED ISSUE #41: Add animation in Installments component
Demo by @hannah-ahl  (slowed down by the gif converter)

installments-animation

HTML + CSS demo:
[Toolkit-Installment-Selector-20160902T112128Z.zip](https://github.com/klarna/ui/files/451886/Toolkit-Installment-Selector-20160902T112128Z.zip)

Nevon:
animated-installments

Mostly works in horizontal mode now. The borders don't re-appear at the
correct time. I think maybe I'll have to use animation instead of
transitions for those.

Haven't started working on vertical mode yet. It becomes a lot trickier,
since the vertical height is not the same for all the options.

Fixes #41

### DIFF
--- a/Installments/index.jsx
+++ b/Installments/index.jsx
@@ -70,12 +70,14 @@ const Installments = React.createClass({
     return {
       hover: undefined,
       previouslySelected: undefined,
-      highlightPosition: {
-        width: undefined,
-        height: undefined,
-        left: undefined,
-        top: undefined,
-        show: false
+      highlight: {
+        position: {
+          width: undefined,
+          height: undefined,
+          left: undefined,
+          top: undefined
+        },
+        transitions: false
       }
     }
   },
@@ -106,9 +108,9 @@ const Installments = React.createClass({
 
       this.setHighlightPosition(calculateHighlightPosition(label))
 
-      if (this.state.highlightPosition.transitionsEnabled !== true) {
+      if (this.state.highlight.transitions !== true) {
         setTimeout(() => {
-          this.setHighlightPosition({ transitionsEnabled: true })
+          this.setHighlightTransitions(true)
         }, TRANSITION_DURATION)
       }
     }
@@ -133,9 +135,21 @@ const Installments = React.createClass({
 
   setHighlightPosition (position) {
     this.setState({
-      highlightPosition: {
-        ...this.state.highlightPosition,
-        ...position
+      highlight: {
+        ...this.state.highlight,
+        position: {
+          ...this.state.highlight.position,
+          ...position
+        }
+      }
+    })
+  },
+
+  setHighlightTransitions (enabled) {
+    this.setState({
+      highlight: {
+        ...this.state.highlight,
+        transitions: enabled
       }
     })
   },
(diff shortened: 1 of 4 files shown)


xaviervia:
internet-rainbow-meme-11

Nevon:
animated-installments-working mov

Vad tror du kompis, @xaviervia?

xaviervia:
Toppen. Ba' gjört

Nevon:
I've tweaked the animation so it more closely matches the mockup. The easing curve should be the exact same, but I reduced the animation duration from 700ms to 500ms, as 700ms just felt incredibly sluggish.

@hannah-ahl: Think you could take a look at this and see if it's representative of your concept? Note that the gif is slowed down because of gifiness, so it's best to look at it live.

animated-installments-tweaked

Nevon:
I would also appreciate a thorough review by someone. This is the result of a bunch of iterations on different approaches, so it's very likely that there's crap left over that I missed.

hannah-ahl:
@Nevon The installment selector should still have the separating lines, I didn't include them in the mock since I only built it to show the animation (sorry, I should have made that clear when I posted the mock)

Nevon:
@hannah-ahl: Gotcha. Borders are back. There is a corner case where when you transition between two neighboring cells, there is no border between them. The reason is that the cell that is after the currently selected cell should not have a left border (since that's where the highlight is, so we don't want the gray border to stick out of the rounded edges), and neither should the currently selected for the same reason. So if you transition between "after the currently selected" to "selected", the transition is from not having a border to not having a border. Ideally we'd want to have the border in the transition period and then remove it when the transition is over, but I'm not certain that's possible.

animated-installments-tweaked mov

Installments/index.jsx
```diff
@@ -13,6 +13,59 @@ const classes = {
...
+  let waiting = false
+
+  return function (...args) {
+    if (!waiting) {
+      waiting = true
+
+      requestAnimationFrame(() => {
+        fn(...args)
+        waiting = false
+      })
```
joaomilho: We already have react-motion in the build (for the smoooth menus), so why all this effort?

Nevon: I wasn't actually aware that we were using react-motion, since basically nothing in the toolkit is animated at the moment.

The reason I'm messing about with requestAnimationFrame is because we need to reposition the highlight on resize, which is called a million times when you resize the window. Since `calculateHighlightPosition` causes layout due to calling `offsetLeft` etc, it should only be called when we are actually re-rendering.

joaomilho: this --> https://github.com/klarna/ui/blob/master/Menu/Tab/index.jsx

joaomilho: to be more specific --> https://github.com/klarna/ui/blob/master/Menu/Tab/index.jsx#L102-L104

Nevon: Yeah, I know what you mean, but the problem is not doing the motion, it's setting the state on resize. We need to recalculate the position on resize and store that in the state. This is not the position and dimensions while in transition, but the final position and dimensions after the transition has completed. If we try to do that on every resize, we'll completely kill performance, since we have to layout in order to find that position.

PR #146, opened 2016-10-24
Radio button tweaks

pirelenito:
This PR does two changes:

- Removes the hover state that would highlight the backgroud as gray;
- Add a new option `borderfull` to add borders between options.


pirelenito:
I'm just fixing the grid.... hang tight!

pirelenito:
Ready for review! The grid has been restored.

pirelenito:
Here is how it looks:

screen shot 2016-10-24 at 16 40 36

PR #147, opened 2016-10-25
Add default name to form components

### LINKED ISSUE #67: Add uuid names to components when not provided
This is so controlled and uncontrolled components that need names to operate such as the Switch, Radio and so on can still work in case a name is not actually passed.

The ids should be built in a clever way so that:
- If actually submitted, it can be tracked down at least what _type_ of element was the one doing the submit (for example, prefixing the uuid with the type, such as `Field-<uuid>`).
- The uuids are not regenerated on each re render. For this purpose a counter in the module closure could be use.

Nevon:
Fixes #67 

So this is a first attempt at generating stable, unique names for form controls where the name is not specified as a prop.

I've only added it to the uncontrolled field at the moment, so to try it out, remove the `name` from one of the examples where it is used. When the control re-renders, the name should stay the same, and the name should be unique.

The general idea is that we concatenate the component name with a UUIDv4, which is seeded using a counter in the closure. This means that the uuid will very likely be re-used, but only with other components. There should never be a complete name collision due to the component name being different.


xaviervia:
OK, finally got the chance to go back to this one. 

Love it. Perfect approach. I will merge the "make everything uncontrolled" branch into this and put the higher-order component in all the relevant components, I think we’ll be good to go.

xaviervia:
Works now. @Nevon mind to review?

lib/uuid.js
```diff
@@ -0,0 +1,13 @@
+import seededRandom from 'seed-random'
```
Nevon: This shouldn't be needed anymore, right, since you're using the higher-order-components package instead. Same for `lib/uniqueComponentName`.

Nevon:
And now the build is failing because there's an imprecise dependency of react. react-maskedinput seems to be importing a private module from react, which either no longer exists or has been moved. https://github.com/insin/react-maskedinput/issues/76

PR #150, opened 2016-10-26
Revamped radio selector

pirelenito:
This PR normalizes how we do alignment inside the Radio options. 
- The checkbox itself is vertically aligned with the content
- The top and bottom margin are now consistent (which was a problem when `borderful` was selected)

Here is how it looks now in a wide range of use-cases:

screen shot 2016-10-26 at 14 46 25

### DIFF
--- a/Collapsible/index.jsx
+++ b/Collapsible/index.jsx
@@ -34,7 +34,7 @@ export default class Collapsible extends Component {
   render () {
     const {children, collapsed} = this.props
 
-    return <div ref={(div) => {this.content = div}}>
+    return <div ref={(div) => { this.content = div }}>
       <Motion style={{height: spring(collapsed ? 0 : this.state.height)}}>
         {({height}) => <div style={{height: height, overflow: 'hidden'}}>
           {children}
--- a/Collapsible/index.jsx
+++ b/Collapsible/index.jsx
@@ -35,8 +35,16 @@ export default class Collapsible extends Component {
     const {children, collapsed} = this.props
 
     return <div ref={(div) => { this.content = div }}>
-      <Motion style={{height: spring(collapsed ? 0 : this.state.height)}}>
-        {({height}) => <div style={{height: height, overflow: 'hidden'}}>
+      <Motion style={{
+        height: spring(collapsed ? 0 : this.state.height),
+        opacity: spring(collapsed ? 0 : 1)
+      }}>
+        {({height, opacity}) => <div
+          style={{
+            height,
+            opacity,
+            overflow: 'hidden'
+          }}>
           {children}
         </div>}
       </Motion>
--- a/Collapsible/index.jsx
+++ b/Collapsible/index.jsx
@@ -1,0 +1,47 @@
+import React, {Component} from 'react'
+import {Motion, spring} from 'react-motion'
+
+export default class Collapsible extends Component {
+  constructor () {
+    super()
+
+    this.state = {
+      height: 0
+    }
+  }
+
+  componentDidMount () {
+    this.setState({
+      height: getHeight(this.content)
+    })
+
+    window.addEventListener('resize', () => {
+      const height = getHeight(this.content)
+
+      if (this.state.height !== height) {
+        this.setState({ height })
+      }
+    })
+  }
+
+  componentDidUpdate () {
+    const height = getHeight(this.content)
+    if (this.state.height !== height) {
+      this.setState({ height })
+    }
+  }
+
+  render () {
+    const {children, collapsed} = this.props
+
+    return <div ref={(div) => {this.content = div}}>
+      <Motion style={{height: spring(collapsed ? 0 : this.state.height)}}>
+        {({height}) => <div style={{height: height, overflow: 'hidden'}}>
+          {children}
+        </div>}
+      </Motion>
+    </div>
+  }
+}
+
+const getHeight = (node) => node.children[0].children[0].getBoundingClientRect().height
--- a/Radio/example.jsx
+++ b/Radio/example.jsx
@@ -1,11 +1,25 @@
 import React from 'react'
 import Radio from '../Radio'
 import UncontrolledRadio from '../uncontrolled/Radio'
-import { LIVE } from '../Showroom/variationTypes'
+import { LIVE_WIDE } from '../Showroom/variationTypes'
+import Installments from '../Installments'
 
 const options = [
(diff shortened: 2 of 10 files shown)


pirelenito:
Currently adding support to expandable content... Hang tight!

xaviervia:
so-smoooth

Collapsible/index.jsx
```diff
@@ -0,0 +1,47 @@
...
+  componentDidMount () {
+    this.setState({
+      height: getHeight(this.content)
+    })
+
+    window.addEventListener('resize', () => {
+      const height = getHeight(this.content)
+
+      if (this.state.height !== height) {
+        this.setState({ height })
```
Nevon: This is going to cause the `render` method to get called on every resize event, which you might have happen more than once per actual browser rendering. Since `getBoundingClientRect` causes a layout, this is very expensive. Check #144 to see how I debounced the event so it only happens when the browser will actually render a new frame. It made a world of difference in some browsers.

xaviervia:
After we talked with @omriyariv and got feedback from @benklarna , the `borderful` should be the default, so let's change that.

xaviervia:
Borderful is now the default.

Question for @hannah-ahl, @benklarna , are we ever using the `borderless` version? I mean this:
screen shot 2016-10-27 at 10 45 11

Collapsible/index.jsx
xaviervia: 👍

xaviervia: Just imagine this with your animated Installments inside

xaviervia:
Add one element disabled option
screen shot 2016-10-28 at 16 02 48

Collapsible/index.jsx
xaviervia: I don't think we have time to fix this before the current deadline. I'll move your comment to it's own issue since I guess this is not a showstopper.

Radio/example.jsx
```diff
@@ -1,12 +1,75 @@
...
 
 const options = [
   {key: 'lorem', label: 'Lorem', description: 'Lorem Ipsum is simply dummy text of the printing and typesetting industry.'},
   {key: 'sit', label: 'Sit', description: 'Amet et consequetur'},
-  {key: 'ipsum', label: 'Ipsum', description: "Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book."}
+  {key: 'ipsum', label: 'Ipsum', description: 'Dummy text ever since the 1500s.'}
+]
+
+const card = <svg width='32px' height='22px' viewBox='0 0 32 22'>
+  <g stroke='none' stroke-width='1' fill='none' fill-rule='evenodd'>
```
JoakimLofgren: Isn't it `strokeWidth` and `fillRule`?

Showroom/examples/compositions/UncontrolledForm.jsx
```diff
@@ -7,9 +7,69 @@ import * as Button from '../../../Button'
...
+import * as Checklist from '../../../Checklist'
 import {SHOWCASE} from '../../variationTypes'
 
+const optionsWithContent = [
+  {
+    key: 'lorem',
+    label: 'Lorem',
+    description: 'Lorem Ipsum is simply dummy.',
+    aside: <svg width='32px' height='22px' viewBox='0 0 32 22'>
+      <g stroke='none' stroke-width='1' fill='none' fill-rule='evenodd'>
```
JoakimLofgren: Should be strokeWidth etc here as well

Radio/index.jsx
```diff
@@ -67,19 +84,42 @@ export default React.createClass({
...
+            label,
+            description,
+            disabled: toExtractDisabled, // eslint-disable-line no-unused-vars
+            aside,
+            content,
+            leftPad,
+            ...restOfProps
+          } = option
+
+          const disabled = allDisabled || option.disabled
```
JoakimLofgren: Why are you destructing `disabled` to `toExtractDisabled` if you still refer to it as `option.disabled` here?

Collapsible/index.jsx
```diff
@@ -0,0 +1,62 @@
...
+            overflow: 'hidden'
+          }}>
+          {children}
+        </div>}
+      </Motion>
+    </div>
+  }
+}
+
+const getHeight = (node) => node.children[0].children[0].getBoundingClientRect().height
```
JoakimLofgren: Don't you do `getBoundingClientRect()` on the div on line 49? So why not ref to it directly?

Collapsible/index.jsx
```diff
@@ -0,0 +1,62 @@
...
+    super()
+
+    this.state = {
+      height: 0
+    }
+  }
+
+  componentDidMount () {
+    this.debouncedResizeHandler = debounce(() => this.onResize())
+    window.addEventListener('resize', this.debouncedResizeHandler)
```
JoakimLofgren: What is the handling of resize events actually needed for ?
Does the height change for the element(s) on resize ?

Radio/example.jsx
```diff
@@ -15,64 +78,90 @@ export default {
   examples: {
     require: `import Radio from '@klarna/ui/Radio'
 import UncontrolledRadio from '@klarna/ui/uncontrolled/Radio'`,
-    type: LIVE,
+    type: LIVE_WIDE,
 
     examples: {
       Regular: <Radio
+        autoFocus={false}
```
JoakimLofgren: Is this needed? Isn't it equivalent to not setting autoFocus ?

Collapsible/index.jsx
Nevon: Yes. The installments component, for example, becomes a vertical list rather than a horizontal list at a certain breakpoint. But it could also just be as simple as a line of text breaking into two lines.

Radio/index.jsx
Nevon: Fixed by 56d081f

Showroom/examples/compositions/UncontrolledForm.jsx
Nevon: Fixed in e88c891

Radio/example.jsx
Nevon: Fixed in e88c891

Collapsible/index.jsx
Nevon: I don't know, but I tried it and it didn't work. I haven't looked into what dark magic react-motion is doing.

pirelenito:
OMG! 😭

pirelenito:
Btw, this should be a major release. So lets hold the merge while we decide how to do it properly.

PR #154, opened 2016-10-31
Create static autoprefixing list

### LINKED ISSUE #153: No transition in installments in old iOS
I guess we're missing a vendor prefix somewhere.

Nevon:
The default autoprefixer configuration is `'> 1%, last 2 versions, Firefox ESR`. We need support for ie8+ and ios safari 8+, so rather than dynamically getting the configuration, I created a static configuration that is based on `browserslist '> 1%, last 4 versions, Firefox ESR, ie >= 8, ios_saf >= 8, safari >= 8'`.

ios-8-installments mov

Fixes #153


Nevon:
Got a 👍 from @xaviervia who is too lazy to open up his laptop.

PR #155, opened 2016-10-31
Remove min with on insallments

martinnorum:


Nevon:
For the sake of future software archeologists, the reason this is being removed is because in KCO we are putting the installment options within a radio selector, which means that it could be < 280 px.

PR #157, opened 2016-10-31
Add CodePrompt template

xaviervia:
screen shot 2016-11-03 at 15 57 45
screen shot 2016-11-03 at 15 58 11
screen shot 2016-11-03 at 16 03 50


xaviervia:
Add `length` prop to set up how long should the "input" be.

xaviervia:
Note: the spacing between the title and the subtitle is 1 grid space less than in the sketch files, but this is the way it looks in the other templates.

screen shot 2016-11-03 at 13 06 53

xaviervia:
- [ ] Make overridable
- [ ] Make themeable

PR #158, opened 2016-10-31
Manually handle hover state in Installments

### LINKED ISSUE #156: Installments receive focus styling on touch
When scrolling in iOS 8 or 9, the installment options receive focus when touched, and retain it after the touch ends.

To reproduce, go to the showroom and scroll down, making sure that you touch one of the installment options as you do so. The option won't be selected, but the "hover" will be sticky.

martinnorum:
Fixes #156

### DIFF
--- a/Installments/index.jsx
+++ b/Installments/index.jsx
@@ -219,6 +219,7 @@ const Installments = React.createClass({
               classes.cell,
               { 'is-focused': focus === key },
               { 'is-selected': key === selected },
+              { 'is-hovered': id === this.state.hover },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@klarna/ui",
-  "version": "1.17.2",
+  "version": "1.17.3-rc.0",
```
xaviervia: Oh we don't actually do this. Versions should be updated in master with `npm version` so that tags are added and so on.

xaviervia:
Other than that small change in the version, LGTM.

package.json
martinnorum: Cool. We've removed that commit now.

PR #159, opened 2016-10-31
React and ReactMotion should be peerDependencies

pirelenito:
Otherwise we might bundle two distinct versions of React due to the way npm installs dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,9 +20,7 @@
   "dependencies": {
     "classnames": "2.2.5",
     "parse-color": "1.0.0",
-    "react": "^15.0.2",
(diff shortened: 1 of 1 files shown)


pirelenito:
I guess the idea is that any "big enough" library should be turned into a `peerDependency`.

pirelenito:
React was introduced as a dependency recently at https://github.com/klarna/ui/pull/139/files

joaomilho:
Yes, I removed react as a peerDependency so if one uses Preact he doesn't even need to install it... but if you know this is _really_ needed, ...

PR #161, opened 2016-10-31
Update check checklist style

beveryday:
We have an issue with checklist... When the text is too long the check shows at the bottom of the last line of text.

### DIFF
--- a/Checklist/styles.scss
+++ b/Checklist/styles.scss
@@ -42,7 +42,7 @@
     stroke: map-get($colors, klarna-blue);
     stroke-linecap: round;
     stroke-width: 2;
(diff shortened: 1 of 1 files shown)


beveryday:
screen shot 2016-10-31 at 3 41 31 pm

xaviervia:
Simply setting vertically aligned is misaligned from the grid. We will have to compensate with padding on top of each item.

screen shot 2016-11-02 at 15 30 30

beveryday:
I'm not able to reproduce the look you have there. I did bump the top pixels down to 2px from 5px and that seemed to better align on the checkout.

PR #165, opened 2016-11-02
Make components overridable

Deschtex:
In order for us to be able to override certain components' styles, or even override the entire component, we need to introduce a new `overridable` decorator. It does assume a couple of things, namely that a `design` object with a `getOverrideFor` method exists in the context.

The `getOverrideFor` method is expected to take a component and return an object with `css` and `Component`, where the `css` property is a CSS modules object and the `Component` is a React component.

This will be a breaking change if there are people using the `themeable` decorator directly in their projects.

Also, I only added it to the currently themeable components.

### DIFF
--- a/decorators/overridable.jsx
+++ b/decorators/overridable.jsx
@@ -6,7 +6,7 @@ const overridable = (styles = {}, designName) => (Target) => {
     constructor (props) {
       super(props)
 
-      this.designName = designName ||  Target.displayName || Target.name
+      this.designName = designName || Target.displayName || Target.name
       this.Component = Target
     }
     componentWillMount () {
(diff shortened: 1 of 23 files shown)


xaviervia:
LGTM

decorators/overridable.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { withPropsFromContext } from 'react-context-props'
+
+const overridable = (styles = {}, designName) => (Target) => {
+  const OverridableComponent = React.createClass({
```
joaomilho: Can we avoid using `createClass` here? I think it's not compatible w/ Preact and it may be deprecd soon.

Deschtex: What should we use instead? We use `React.createClass` for all other components.

Deschtex: I changed to using a ES6 class instead

PR #168, opened 2016-11-03
Disable installments cell highlight evens

npejo:
Disable mouse interactions with the cell highlight element. The absolutely positioned highlight cell is preventing mouse interaction with the label element which is below it.


jonotrujillo:
Looks good!

PR #177, opened 2016-11-08
Add loading-state for the landing page

cskeppstedt:
We need this to avoid clicking the button twice (clicking the button performs an HTTP POST in the authentication UI).

landing-page-loading-state


cskeppstedt:
In general, the buttons will always resize when they transition into the loading state because of the fact that they don't have 100% width anymore. I think that could be avoided by keeping the text with `display: none` and position the spinner with absolute position, if desired.

Issue #170, opened 2016-11-04
Scope down CSS transitions


xaviervia:
Right now almost of components specify `transition: all .2s ease`. They should be more specific about the target CSS properties to transition.

Deschtex:
Yes, yes, yes

PR #184, opened 2016-11-10
Scale down css transitions

xaviervia:
Solves #170 

Hopefully this will mean that no ugly flashes of "appear" animations will happen anymore.

@Deschtex skulle du vilja recensera?


xaviervia:
OK, I'll merge this since it's pretty safe either way.

PR #188, opened 2016-11-16
Trigger blur when PinCode field is disabled

cskeppstedt:
Some devices (iOS 10) will leave the keyboard open if the field is disabled while it has focus. Since `PinCode` is used as an auto-submitting field, this will always happen.

I had another version where the logic was a more verbose and explanatory, like:

```jsx
    componentWillReceiveProps (nextProps) {
      const inputExists = !!this._input
      const hasFocus = this.state.focus
      const willBeDisabled = nextProps.disabled && !this.props.disabled

      if (inputExists && hasFocus && willBeDisabled) {
        this._input.blur()
      }
    }
```

Not sure if it adds that much though. The code as it stands now, along with the git log message, should be enough to explain what's going on. I think.

### DIFF
--- a/Field/PinCode/index.jsx
+++ b/Field/PinCode/index.jsx
@@ -18,6 +18,12 @@ class PinCode extends PureComponent {
     }
   }
 
+  componentWillReceiveProps (nextProps) {
+    if (nextProps.disabled && !this.props.disabled && this.state.focus && this._input) {
+      this._input.blur()
+    }
+  }
+
   render () {
     const {
       className,
@@ -55,6 +61,7 @@ class PinCode extends PureComponent {
       onMouseEnter={() => this.setState({hover: true})}
       onMouseLeave={() => this.setState({hover: false})}
       maxLength={length}
+      ref={(elem) => this._input = elem}
       style={inputStyle}
       type='tel'
       value={value}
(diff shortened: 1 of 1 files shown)


Nevon:
I actually prefer the verbose version, and would even like it if the entire implementation would be moved into a `blurOnDisableForIOS10` function that's called from `componentWillReceiveProps`, just to make it obvious _why_ it's being done, rather than _what_ is being done.

Other than that, it looks good to me.

cskeppstedt:
If we'd call it `blurOnDisableForIOS10` I think we should explicitly target iOS10 though. I'm not sure if this happens for other iOS versions, but possibly, I can check. How about `blurOnDisable`? Also the API of the function is kind of funky, imagine:

```jsx
componentWillReceiveProps (nextProps) {
  blurOnDisable (nextProps.disabled, this.props.disabled, this.state.focus, this._input)
}
```

There are other ways of doing it of course, right now I can't come up with one that looks great. Let me think for a minute :)

Nevon:
Maybe `blurOnDisable(input)(nextProps, props, state)` is a more reasonable signature. Not sure. It still doesn't communicate the _why_, which is my main issue. Maybe just a comment in the function body.

cskeppstedt:
@Nevon wdyt? 210cb9b

Nevon:
Lovely! Especially the commit message.

PR #189, opened 2016-11-16
Ensure PinCode input loses focus when it is disabled

MikaelAmborn:
This is to make sure the keyboard on iOS gets hidden when to input field
gets disabled.


MikaelAmborn:
Duplicate of #188

PR #190, opened 2016-11-16
Separate error state from error message in CodePrompt

cskeppstedt:
We have a use case in the Authentication UI to show an error message, but the input should not be in the error state. This was our take on how to accomplish that.


Nevon:
How about defaulting `error` to true if there is an `errorMessage`? It seems to me that showing an error message without being in an error state is a corner case, so maybe it makes sense to have to be explicit about the fact that it should not be in an error state?

cskeppstedt:
@Nevon in case `error` is undefined you mean? That could make sense.

Nevon:
Yes, exactly.

cskeppstedt:
@Nevon funky fresh 27946e1

Nevon:
LGTM!

PR #191, opened 2016-11-16
Include the for label for the radio, only when not disabled

farfalena:


Nevon:
Given the premise that a disabled radio button doesn't have an input element, this looks fine, but just for my own understanding, why is there no input element when it's disabled? Seems like there should be a disabled input element in that case, but maybe I'm missing something.

xaviervia:
@Nevon when I originally removed the input it seemed like a very safe way of disabling interaction with the disabled options. I didn't check if the `disabled` state in an input of type `radio` affects that input alone or the whole radio input set. I guess we can check that. 

I am not sure if this changes this PR in any way though, considering that removing the `for` attribute is probably a good idea even if we add back the input as disabled. Should we make other issue for it?

Nevon:
`disabled` on an input certainly only affects the input that it is applied to, not the entire set (https://jsfiddle.net/4tuvq2yp/), but sure, I'll create a separate issue for that.

xaviervia:
Why did I close this? I think it was an accident. I’ll merge it now along @Nevon’s fix

Nevon:
What happened was that I included a smart commit in #202 that closed this issue without merging, since my PR superseded it. This should not have been merged. Will be fixed by #204

PR #193, opened 2016-11-16
Fix the isDigit function

cskeppstedt:
It wasn't actually using the variable `x`, it would short-circuit to `false || '0'` and return true every time.

@xaviervia WHERE'S YOUR IVORY TOWER NOW? :)

### DIFF
--- a/templates/CodePrompt/index.jsx
+++ b/templates/CodePrompt/index.jsx
@@ -22,12 +22,11 @@ const classes = {
   message: `${baseClass}__message`
 }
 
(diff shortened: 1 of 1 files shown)


Nevon:
Ivory tower version: `const isNumber = (x) => Array.apply(null, {length: 10}).map(Number.call, Number).map(x => x.toString()).includes(x)`

cskeppstedt:
Yeah we were just parsing that. I think we'll stick with our version for now, the keeper of the ivory tower may introduce the purer version later :)

PR #199, opened 2016-11-17
Fix/pincode in safari

xaviervia:
Sorry no screenshots yet


cskeppstedt:
Lgtm, ship it if you agree with my addition :)

PR #201, opened 2016-11-18
Fix spacing in templates

xaviervia:
## Generic dialog changes

- Reduce top and bottom visual margins to 50px for non textual components (buttons, dropdowns, illustrations) and to 60 for textual components. This was agreed with @hannah-ahl @zenona 

## Centered chrome changes

- Reduce spacings between parts of the Centered chrome (title with summary, summary with button, etc)
- Set a maximum width on the Title and center it (we will need to keep the copy arranged so that it folds nicely)

## Added a Strong component

The reason why this is important is that Open Sans font weight for bold is actually `600`, not `700`, so we need to keep consistency.

screen shot 2016-11-18 at 13 50 37


Nevon:
I'm a little confused about the Strong component. Why is it necessary?

xaviervia:
@Nevon the Open Sans semi bold. People doing `html2react` or markdown or something like that need to adjust `strong` markup to use the Open Sans bold style that we use, which is `600` font weight, not `700`. But Strong is also a semantically relevant feature, so style overrides need to be able to target it: right now we are doing it via overriding the `strong` tag, but this would be much cleaner.

Nevon:
Gotcha!

Issue #171, opened 2016-11-04
Improve IconButtons


xaviervia:
- [x] Add displayName `IconButton` as a namespace to the components
- [ ] Change the class names to better reflect the structure (no `illustration`, just `icon-button`)

xaviervia:
Irrelevant when we do the switch to inline styles.

PR #206, opened 2016-11-22
Add displayName to IconButtons

Nevon:
Contributes to #171


xaviervia:
So elegant. Should we do it in https://github.com/klarna/higher-order-components as well?

Nevon:
Sure. I'm moving it in there now.

Issue #181, opened 2016-11-10
Footprint


joaomilho:
Following @ctaintor 's priorities, I feel we could find a way to estimate and present both component's and the whole lib size in the showcase, and/or as a test, and/or as a build step. The goal is mainly to create awareness and make us focus on leaner solutions.

JoakimLofgren:
We could create a little graph over all the versions to see how the size varies.

Nevon:
I recently learned about a package called `source-map-explorer`, which is the best bundle analyser I've seen so far.

I ran it against a build of the latest master. You can see the interactive results here: https://rawgit.com/Nevon/22c1f6a5b0aa5bf7c8350248c00318e3/raw/ec408f51a1c7834201eb646ef15680d14ea25162/klarna-ui-bundle-analyze.html

Not terribly useful on a day to day basis, but it could be useful to explain sudden increases in bundle size, to see that you're not bundling something twice or including a dependency that's much larger than you'd have thought.

xaviervia:
Well @Nevon 's solution is in master now! I'm closing this one :)

PR #207, opened 2016-11-22
Track bundle size changes

Nevon:
As a first step for #181, it should be simple enough to at least show how a PR is affecting the bundle size.

Given that Travis executes the prepublish step and builds the distributed version, this should work.

Hopefully we'll see something like this when this PR is built, though there's no built-index.js to compare to, so we'll just see the absolute size, I think. But once it's in place, it'll compare it to the target branch and show you the relative difference.

Example

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ before_script:
 - npm install
 script:
 - npm test
+- curl -O https://raw.githubusercontent.com/danvk/travis-weigh-in/master/weigh_in.py && python weigh_in.py index-built.js
 deploy:
   provider: npm
   email: ui-admin.e@klarna.com


Nevon:
So I had to give it a personal access token, so it's posting as me right now. 💪 Not sure if it's possible to get the token that travis is using instead.

Nevon:
Alright, so I felt a little uncomfortable with having a personal access token available for just about anyone to see, so I created what Github calls a "machine user" (meaning literally just an alternative account), gave it access to this repo and created a personal access token for it. Say hello, @TravisWeighIn!

TravisWeighIn:
Yo! That's a nice bundle you got there!

xaviervia:
sentry

PR #212, opened 2016-11-23
Have the Dialog respect the custom styled border-radius

### LINKED ISSUE #211: Dialog should support custom styling, like borderRadius

farfalena:
Fixes #211 @xaviervia

### DIFF
--- a/Dialog/index.jsx
+++ b/Dialog/index.jsx
@@ -15,11 +15,21 @@ const classes = {
   table: `${baseClass}__table`
 }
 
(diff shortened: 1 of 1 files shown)


xaviervia:
Looks good. We need to make it also `themeable` for it to be able to pick up the `customizations` from the context. Something like this should do: https://github.com/klarna/ui/blob/master/Checklist/index.jsx#L96

farfalena:
@xaviervia, thanks for noting that, addressed in b77155d.

PR #214, opened 2016-11-24
Do not promote custom styling for Text elements, when the color prop is error or warning.

### LINKED ISSUE #213: Paragraph.Primary should not be overriden with custom styles in case of error or warning

farfalena:
Fixes #213 @xaviervia

### DIFF
--- a/Paragraph/Primary/index.jsx
+++ b/Paragraph/Primary/index.jsx
@@ -5,6 +5,7 @@ import overridable from '../../decorators/overridable'
 import defaultStyles from './styles.scss'
 import palette from '../../lib/palette'
 import compose from '../../lib/compose'
+import { shouldTheme } from '../../Text'
 
 const baseClass = 'paragraph--primary'
 
@@ -54,16 +55,13 @@ Primary.propTypes = {
 }
 
 export default compose(
-  themeable((customizations, props) => {
-    if (['error', 'warning'].includes(props.color)) {
-      return {}
-    }
-    return {
+  themeable((customizations, props) => (
+    shouldTheme(props) ? {
       style: {
         ...props.style,
-        color: customizations.color_text
+        color: customizations.color_header
       }
-    }
-  }),
+    } : {}
+  )),
   overridable(defaultStyles)
 )(Primary)
--- a/Paragraph/Primary/index.jsx
+++ b/Paragraph/Primary/index.jsx
@@ -54,11 +54,16 @@ Primary.propTypes = {
 }
 
 export default compose(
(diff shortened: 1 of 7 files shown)


Paragraph/Primary/index.jsx
```diff
@@ -54,11 +54,16 @@ Primary.propTypes = {
 }
 
 export default compose(
-  themeable((customizations, props) => ({
-    style: {
-      ...props.style,
-      color: customizations.color_text
+  themeable((customizations, props) => {
+    if (['error', 'warning'].includes(props.color)) {
```
Nevon: I really don't like this multiple return inside a conditional, but that's my personal opinion, so do with it what you will.

But rather than repeating this logic across multiple components, couldn't you do something like:

```
// themeableDefaultState.js
const UNTHEMED_STATES = ['error', 'warning']
const noTheme = function () { return {} }

const shouldTheme = function ({ color }) {
  return !UNTHEMED_STATES.includes(color)
}

export default function (adapter, props) {
  return themeable(shouldTheme(props) ? adapter : noTheme)
}

// Paragraph/Primary/index.jsx

export default compose(
  themeableDefaultState(function (customizations, props) {
    return {
      style: {
        ...props.style,
        color: customizations.color_text
      }
    }
  }, props),
  overridable(defaultStyles)
)(Primary)
```

Alternatively, if you think that the logic will actually be different between different components (maybe the prop to look at isn't `color`, or maybe the unthemed states are different, maybe just pull out the `UNTHEMED_STATES.includes(props.color)` part into a private function in the components, so that it can be made a little bit more explicit _why_ this logic is there and what it means.

farfalena: Some reuse should definitely be done, I will look into it. However I think that we cannot avoid having the check for whether to theme or not inside the `adapter` function, given that it is there were we have access to the `props`...

farfalena: Defined and reused a `shouldTheme` function for all Text components in 062dcbb

Nevon: Yeah, you're right that the distinction needs to be done in the adapter. My bad. But we could still have a function that takes an adapter and returns a new adapter where the distinction can be done. Something like:

```
themeable(ifThemeableState(() =>({
  style: {
    ...props.style,
    color: customizations.color_header
  }
})))
```

Where `ifThemeableState` is:

```
export default (adapter) => (props) => ['error', 'warning'].includes(props.color) ? {} : adapter(props)
```

farfalena: Addressed in c2f6c74

PR #215, opened 2016-11-26
Remove all unneccessary arrow functions

Nevon:
Trying this out to see if there is any noticable change in the bundle size. I've only removed the ones that actually get built into the bundle.

This _should_  be functionally equivalent, but it's very possible that I've made a mistake, as this was the most boring refactor I've ever done. Mostly this is an experiment.


Nevon:
Hah, -39 bytes 🎉

PR #216, opened 2016-11-27
Fix the Installments component growing on iphone6 landscape

npejo:
Safari
screen shot 2016-11-27 at 07 29 58

Firefox
screen shot 2016-11-27 at 07 31 32

Chrome
screen shot 2016-11-27 at 07 32 17

IE10
screen shot 2016-11-27 at 07 33 27

### DIFF
--- a/Installments/index.jsx
+++ b/Installments/index.jsx
@@ -32,10 +32,10 @@ const calculateHighlightPosition = (selected) => {
 
   const BORDER_SIZE = 1
 
-  const left = selected.offsetLeft
-  const top = selected.offsetTop
-  const width = selected.offsetWidth + BORDER_SIZE * 2
-  const height = selected.offsetHeight + BORDER_SIZE * 2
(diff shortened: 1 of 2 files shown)


Nevon:
Looks good, but do you think you could write an explanation of what the problem was and how your changes addresses it? It's a little difficult to get from the diff.

npejo:
This PR should fix the following issue that the width of the iframe expands when you select the last option on iPhone6 in landscape orientation.
ever-expanding-width

At the current solution we are setting the rounding border to the parent element and only the separators are borders of the items. When we calculate the position for the highlight element for the last option, we push it most right. On mobile safari, the element is pushed 1px more to the right (outside of the parent element) and because the parent has 100%, it grows to follow the dynamic element. That triggers full re-render of the component and this is how the ever expanding effect happens.

In this fix the border of the parent is removed and it is applied appropriately to each of the options and the border style changes depending on the selected option.

PR #217, opened 2016-11-30
Add handler for orientationchange to installments component

npejo:
If you select the 4th payment option on iphone 6 in landscape mode and then rotate the phone, the component will end up in a state like on the following image

img_5436

The position of the highlight element is not recalculated.

This PR is adding handler on `orientationchange` which updates the position to the highlight element to 0, 0 and that will trigger recalculation.

### DIFF
--- a/Installments/index.jsx
+++ b/Installments/index.jsx
@@ -93,12 +93,15 @@ const Installments = React.createClass({
     }
 
     this.debouncedResizeHandler = debounce(() => this.onResize())
-    window.addEventListener('resize', this.debouncedResizeHandler)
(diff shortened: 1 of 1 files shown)


Installments/index.jsx
```diff
@@ -131,6 +134,15 @@ const Installments = React.createClass({
     this.setHighlightPosition(calculateHighlightPosition(this.getSelectedLabel()))
   },
 
+  onOrientationChange () {
+    this.setHighlightPosition({
```
npejo: The reseting of the highlight position is moved into helper method to a776f7a

PR #218, opened 2016-12-01
[Title: Primary Secondary]Use the correct customization property for text color

farfalena:
When implementing #213 accidentally updated the customisation property controlling the color of text for Primary and Secondary Title. Reverting to the proper one.

@xaviervia, @Nevon


xaviervia:
It’s odd that we were using the `color_text` customization though, but if that’s what we were doing, 👍

PR #221, opened 2016-12-02
Hide and show the highlight element on orientation change

npejo:
Depending on the position of the highlight element it can brake the layout when the orientation is changed (from landscape to portail on iphone6).

To fix this, we will hide the highlight element, then show it and position it in the next thick. This will allow to the browser to update the layout properly and we will position the element in the resulting layout.

Browsers trigger `orientationchange` whenever the height becomes bigger than the width and vice versa. This PR prevent triggering of the `orientationchange` handler on resize of the browser window on desktop. In that case the media queries should handle the layout change.

### DIFF
--- a/Installments/index.jsx
+++ b/Installments/index.jsx
@@ -163,7 +163,7 @@ const Installments = React.createClass({
   },
 
   resetHighlightPosition () {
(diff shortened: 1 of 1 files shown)


Nevon:
This is crazy.

PR #223, opened 2016-12-05
Remove Dialog.Content from templates/Centered

farfalena:
Replace Dialog.Content with a simple div to make the Centered component more reusable across different settings.

@Deschtex, @xaviervia


xaviervia:
As we were discussing in the channel, this implementation is there by
design. Let's discuss alternatives to fix this issue.

PR #224, opened 2016-12-08
Add high brand volume to Buttons and templates

xaviervia:
Templates are pending:

screen shot 2016-12-08 at 16 28 02
screen shot 2016-12-08 at 16 28 07


joaomilho:
"brand volume" means "rounded"?

xaviervia:
Among other things. It means that it’s more branded. What "more branded" will specifically mean will be a different thing and evolve in the future. The idea is that consumer do not have to setup those configurations separately, just setup a "brand volume" and delegate the implementation.

In the end this property will probably be captured from the React.context in a similar fashion as the `<Theme>` component and `themeable` HOC.

PR #232, opened 2016-12-12
Allow content to overflow on an expanded Collapsible

apretto:
Overflow rule to enable content to overflow outside the collapsible once the animation is close to be complete (the last few pixels take a while to be expanded).

'10' is a magic number 🎩


xaviervia:
You guys do your thing

PR #234, opened 2016-12-12
KCOC-1407 - Logos should be "grayed-out"

beveryday:
Logos should be grayed-out for the locked payment options.

### DIFF
--- a/Radio/styles.scss
+++ b/Radio/styles.scss
@@ -100,7 +100,7 @@
   vertical-align: top;
   
   .is-disabled & {
-    opacity: 0.2;
+    opacity: $disabled-selector-logo;
   }
 }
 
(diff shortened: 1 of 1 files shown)


Mellbourn:
There are other is-disabled styles in that file, and they don't use opacity. Will the different disabled elements look equally "grayed-out"?
Is using opacity 0.2 common in the application? In that case, shouldn't 0.2 be a constant?

xaviervia:
Good points. I guess the `0.2` opacity should be a constant. I understand that the logos, being colorful, would probably need opacity as the way of "disabling them".

@beveryday I think we can move the `0.2` as a constant to [settings](https://github.com/klarna/ui/blob/master/css/settings.scss) and merge this one

beveryday:
Done, $disabled-selector-logo

Nevon:
The build is failing.

Issue #236, opened 2016-12-13
Use grey text instead of grey lines color in Text and Icons


xaviervia:
The expected color is #8E8E90

PR #237, opened 2016-12-13
Use a darker grey for icons and text

moriaam:


moriaam:
Fixes [#236](https://github.com/klarna/ui/issues/236)

Issue #240, opened 2016-12-13
Radio aside is not right-aligned in IE9 and Android 4 Tablet


farfalena:
![ie9-left-logo](https://cloud.githubusercontent.com/assets/1564560/21160674/fdbc3b1c-c185-11e6-9537-8bab961aad61.png)

PR #241, opened 2016-12-13
Issue 240/radio aside right align

farfalena:
Related to #240 
@xaviervia Don't know if this is an acceptable way to handle this just to address a limitation by older environments (like IE9 and Android 4)


Radio/index.jsx
```diff
@@ -157,7 +158,7 @@ const Radio = React.createClass({
                     </div>
                   </div>}
 
-                  <div className={classNames(classes.optionRight)}>
+                  <div className={aside ? classNames(classes.optionRight) : classNames(classes.optionRight, classes.optionRightmost)}>
```
xaviervia: It’s probably better to write this using the "object" notation for the `classNames` function:

```javascript
className={classNames(
  classes.optionRight, 
  {
    [classes.optionRightmost]: aside != null
  }
)}
```

Not a big deal though

xaviervia:
There seems to be an issue when loading a single option at a time:

screen shot 2016-12-14 at 07 47 34

In this case the bullet will not be shown. We can probably use the `singleOption` flag from  https://github.com/klarna/ui/blob/a56414588200a8bc526383218b90a9c0623292f9/Radio/index.jsx#L148 to change the left padding of the content for this scenario.

Nevon:
@xaviervia: Seems like `calc` doesn't work in conjunction with `display: table` in IE9: http://stackoverflow.com/questions/22776279/ie9-bug-display-table-width-calc

farfalena:
@xaviervia both comments addressed in latest commit.

PR #245, opened 2016-12-15
Fix theme support for instalments border radius

pirelenito:
It was not being applied to the cells.

screen shot 2016-12-15 at 16 23 18


xaviervia:
@Nevon that is actually an unrelated problem. It wasn’t fixed here but it was introduced here either, it is already there.

Nevon:
:(

Nevon:
Created #246 to keep track of that issue.

PR #247, opened 2016-12-15
Fix spacings for Radio selector items

xaviervia:
**Before**

screen shot 2016-12-15 at 16 40 51

**After**

screen shot 2016-12-15 at 16 42 18

### DIFF
--- a/Radio/example.jsx
+++ b/Radio/example.jsx
@@ -26,7 +26,7 @@ const optionsWithContent = [
     label: 'Lorem',
     description: 'Lorem Ipsum is simply dummy.',
     aside: card,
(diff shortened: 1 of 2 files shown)


Radio/example.jsx
```diff
@@ -26,7 +26,7 @@ const optionsWithContent = [
     label: 'Lorem',
     description: 'Lorem Ipsum is simply dummy.',
     aside: card,
-    content: <Paragraph.Secondary condensed>
+    content: <Paragraph.Secondary style={{marginBottom: -10}}>
```
pirelenito: It is unfortunate that we need to do this 😢.

Issue #249, opened 2016-12-19
Part payment option blocks don't use merchant radius_border value


martinnorum:
The new design for the part payment options doesn't make use of the `radius_border merchant` styling value for top-right and the bottom-left corners in vertical layout.

xaviervia:
This has been fixed long ago actually :D

Issue #160, opened 2016-10-31
Line height in installments is too large


Nevon:
<img width="262" alt="screen shot 2016-10-31 at 17 15 32" src="https://cloud.githubusercontent.com/assets/83586/19861657/b183093e-9f8d-11e6-8e54-61beb7215b76.png">

The last option there looks super weird.

@hannah-ahl

xaviervia:
This got fixed already :)

PR #250, opened 2016-12-19
Fix box selector structure

xaviervia:
> Addresses https://github.com/klarna/ui/issues/249 as well
> Addresses https://github.com/klarna/ui/issues/160 as well

Explanation will come later


martinnorum:
@xaviervia I referenced #160 as well in your comment

Nevon:
There's a lot of duplicated code between the two versions, and I'm quite certain that there's a bunch of logic that's no longer needed. For example, the highlight in the horizontal one could be implemented in pure CSS (or close to it at least) and get a well-needed performance boost.

Provided that the API contract stays, though, we could do that in a separate PR, since it will be pretty much a complete rewrite of at least the horizontal one.

xaviervia:
@Nevon well, the new animation is not definitive. It doesn't look too good either. It was just a compromise to get a fix out of the door.

xaviervia:
@Nevon the animation behavior will very likely change. We should sync with @hannah-ahl about that. I removed some of the unused code with @npejo 's blessing, but I'm aware there is more to do. If you are OK with postponing the CSS changes let's merge this now and iterate asap.

farfalena:
I would opt for making this change available and creating a new issue for cleaning up the CSS.

Nevon:
Go for it. I'll create a ticket for revisiting the implementation.

Nevon:
Ticket created at #251

PR #254, opened 2016-12-22
Fix document.activeElement in IE9. Fixes #253

### LINKED ISSUE #253: IE<=9: Safe usage of document.activeElement
In IE<=9, `document.activeElement` will throw `"Unspecified error"` if `document.readyState` is not `complete`.

One option is to mimic Facebook's [getActiveElement()](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/core/dom/getActiveElement.js) method (thanks @npejo).
They surround `document.activeElement` within a try-catch block, but I would suggest just checking against `document.readyState`

In the meanwhile, the following pattern is recommended:
```javascript
if (document.readyState === 'complete') {
  initApp();
} else {
  document.addEventListener('readystatechange', function() {
    if (document.readyState === 'complete') {
      initApp();
    }
  });
}
```

More resources:
- [`document.activeElement` usages in this repo](https://github.com/klarna/ui/search?utf8=%E2%9C%93&q=activeElement)
- [`document.readyState` docs](https://developer.mozilla.org/en/docs/Web/API/Document/readyState)
- [`readystatechange` event docs](https://developer.mozilla.org/en-US/docs/Web/Events/readystatechange)

Nevon:
Haven't actually tested this. Could someone who actually works in front–end and has a VM with IE9 make sure that it works?

### DIFF
--- a/Installments/Horizontal/index.jsx
+++ b/Installments/Horizontal/index.jsx
@@ -5,6 +5,7 @@ import overridable from '../../decorators/overridable'
 import uncontrolled from '../../decorators/uncontrolled'
 import defaultStyles from './styles.scss'
 import compose from '../../lib/compose'
+import getActiveElement from '../../lib/getActiveElement'
 
 const baseClass = 'installments--horizontal'
 
@@ -52,7 +53,7 @@ const Horizontal = React.createClass({
   componentDidMount () {
     if (
       this.props.focus &&
-      document.activeElement !== this.refs[this.props.focus]
+      getActiveElement(document) !== this.refs[this.props.focus]
     ) {
       this.refs[this.props.focus].focus()
     }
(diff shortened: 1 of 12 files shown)


xaviervia:
I'm getting a pretty weird behavior when doing keyboard navigation: I can only move between two of the items available in Radio or Menu.Tab (although it is working for Installments component)

I'll start by checking if this was or not the case before.

xaviervia:
Indeed it wasn't happening before :(

I wonder how are the two things related

xaviervia:
This is what I get now when trying to debug. I added a console log in `componentDidUpdate`. `lorem` is logging of the selection when I clicked with the mouse, `sit` is when I went down with the keyboard. Will keep posting.

screen shot 2017-01-04 at 15 18 15

xaviervia:
I tried the same in `master` and the effect is the same. Now I'm noticing that in master I have to press the arrow several times to move between the items in Radio. Weird. I'll try out other browsers just in case.

xaviervia:
Marked in progress until further investigations.

PR #255, opened 2016-12-22
Improve performance of collapsible animation

pirelenito:
React Motion was dragging IE9 to a hanging performance. CSS Transitions doesn’t work in there, but it is fine to not have transitions in browsers that do not support it.

The animation is not exactly the same, so we need a validation before proceeding.

collapsible-animation

### DIFF
--- a/Collapsible/index.jsx
+++ b/Collapsible/index.jsx
@@ -1,6 +1,6 @@
 import React, {Component} from 'react'
-import {Motion, spring} from 'react-motion'
 import debounce from '../lib/debounce'
+import defaultStyles from './styles.scss'
 
 export default class Collapsible extends Component {
   constructor () {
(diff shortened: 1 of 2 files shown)


jordiparra:
@hannah-ahl is on vacation, but it looks good to me. 👍

hannah-ahl:
Hannah Ahl
Product Designer
Product
Klarna AB
Reg no: 556737-0431

pirelenito:
I've tested this and the previous implementation in an iPad 3 (retina display) and the performance there is also pretty bad.

I understand animating height will be problematic, but could it be something else? I'll leave this PR open to further investigate it.

Nevon:
Transitioning on height is most definitely the cause. Each frame will have a reflow, which will likely affect page height, which causes a resize event to happen, which queries `getBoundingClientRect`, which causes yet another reflow. Since the collapsible likely isn't within an element with position absolute or fixed, the reflow will go all the way to the root, so in a complex tree it won't be fast.

I really don't know a performant way to do this. I've toyed with the idea of having the sections always be the same height, and instead translate them around in order to achieve the illusion of transitioning their height, but in the end we come back to the fact that the parts around the collapsible have to change their position when the collapsible moves, so we're not getting away from a reflow anyway.

Perhaps, if performance is an issue, we could conditionally apply the transition. It could be UA-based, but ideally we would apply the transition based on the actual performance of the device (benchmark within a `requestIdleCallback` maybe?)

pirelenito:
Work here was moved to #258

PR #258, opened 2016-12-28
Optimise Collapsible rendering

pirelenito:
The are two big changes is this PR to improve the overall performance:

- Moved away from React Motion to CSS Transitions;
- Only calculate the height of the content if it is not collapsed or it is about to be expanded.

A caveat from this change is the new animation:

collapsible-animation

### DIFF
--- a/Collapsible/index.jsx
+++ b/Collapsible/index.jsx
@@ -1,6 +1,6 @@
 import React, {Component} from 'react'
-import {Motion, spring} from 'react-motion'
 import debounce from '../lib/debounce'
+import defaultStyles from './styles.scss'
 
 export default class Collapsible extends Component {
   constructor () {
@@ -49,22 +49,15 @@ export default class Collapsible extends Component {
     const {children, collapsed} = this.props
 
     return <div ref={(div) => { this.content = div }}>
-      <Motion style={{
-        height: spring(collapsed ? 0 : this.state.height),
-        opacity: spring(collapsed ? 0 : 1)
-      }}>
-        {({height, opacity}) => <div
-          style={{
-            height,
-            opacity,
-            // Overflow rule to enable content to overflow outside the collapsible
-            // once the animation is close to be complete (the last few pixels take a while
-            // to be expanded). '10' is a magic number 🎩
-            overflow: this.state.height - height > 10 ? 'hidden' : 'visible'
-          }}>
-          {children}
-        </div>}
-      </Motion>
+      <div
+        className={defaultStyles.wrapper}
+        style={{
+          height: collapsed ? 0 : this.state.height,
+          opacity: collapsed ? 0 : 1,
+          overflow: collapsed ? 'hidden' : 'visible'
+        }}>
+        {children}
+      </div>
     </div>
   }
 }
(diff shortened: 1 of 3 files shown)


lib/debounce.js
```diff
@@ -1,16 +1,11 @@
...
-
-  return function (...args) {
-    if (!waiting) {
-      waiting = true
-
-      requestAnimationFrame(() => {
-        fn(...args)
-        waiting = false
-      })
+  return function () {
```
pirelenito: I've replaced the previous faux implementation with a real `debounce`.

Nevon: What was the reasoning behind this? The purpose of the previous implementation was to only invoke the function whenever a paint was about to happen. With the new implementation, you're delaying execution an arbitrary amount of time (possibly indefinitely, if the function is called repeatedly).

pirelenito: Gotcha. We have an application that runs in an iframe that gets resized via an animation. This animation is triggered whenever the Radio selector changes in size (selecting different options); the previous `debounce` would trigger a lot of costly `Collapsible.updateHeight` calls, that is why I changed it.

Should we keep the previous implementation, but rename it to something else (since it is not doing what is expected of a `debounce` function) ? And also keep the new implementation to be used in scenarios such as the one I've described above?

Nevon: Hmm, yeah, I didn't think about the fact that the height of the collapsible is transitioned. My thinking behind coupling it with the paint is that if your application is running slowly, you're automatically going to execute the costly update only when it is necessary (right before painting), but still never end up painting a frame where the height is incorrect. But if the height is transitioned anyway, those in-between frames don't really matter.

If the window is resized 10 times over a period of 500ms, the height of the collapsible will _never_ be resized until the window has finished resizing, as the debounce will delay the initial execution by 300ms, and then keep delaying it every 50ms, which means that the collapsible won't be resized until after 800ms (+ however long it is until the next paint after the timeout finally executes).

I'm not crazy about that delay that will be introduced. If you're resizing the page, there will be a noticable delay until the collapsible starts transitioning. Could you maybe play with the timeout value so that the update is never executed until the resizing is done, but the delay is low enough that it's not so noticable?


Then again, if the height of the collapsible is transitioned anyway, those in-between frames might not matter, since

pirelenito: You were right. We managed to fix the issues by simply being more smart in the Collapsible implementation and preemptively abort calculating not needed sizes.

Collapsible/index.jsx
```diff
@@ -12,8 +12,8 @@ export default class Collapsible extends Component {
   }
```
joaomilho: can be removed?

Collapsible/index.jsx
```diff
@@ -12,8 +12,8 @@ export default class Collapsible extends Component {
...
 
   componentDidMount () {
-    this.debouncedResizeHandler = debounce(() => this.onResize())
-    window.addEventListener('resize', this.debouncedResizeHandler)
+    this.debouncedUpdateHeight = debounce(() => this.updateHeight())
```
joaomilho: this is not really a `debounce` amirite?

pirelenito: Exactly. Should we rename the `debounce` function? wdyt @Nevon ?

Nevon: If you can think of a good name, go for it!

PR #264, opened 2017-01-11
'Add' icon

DonGideon:
tiny new icon of:
screen shot 2017-01-11 at 5 00 10 pm

so nice :) so simple :)

party time_8603f1_4796446


xaviervia:
There is a failing test that is trying to include a file that it doesn’t exist, but other than that it LGTM

DonGideon:
well that should have fix it :)

PR #266, opened 2017-01-11
Disable animation on devices with poor performance

pirelenito:
It uses the module [collect-fps](https://www.npmjs.com/package/collect-fps) to do a health check of the performance of the application and disables animation in the component if the experience is not good enough.

It only applies to the Collapsible component atm, and I’m not sure it sure be merged before a more complete solution is defined.


xaviervia:
Misa likey. But I was part of this so don’t take my word for it.

Nevon:
This is pretty cool, if I'm reading this right, you keep running the profiler every time there is a prop change and animations are enabled, until the device fails to hit the FPS target, at which point you will turn off animations and never run the profiler again.

I guess the only problem I see is that you could have a prop change that is completely unrelated to animations, which takes sufficiently long to complete so that the profiling fails. In that case, animations will be turned off even if it would play flawlessly for the prop changes that actually affect it.

xaviervia:
The check should only run when the `collapse` prop is modified, which always result in some sort of animation. In general though there will not be "false negatives" for 60 fps because of that, since collecting FPS while the browser is not animating results in 60 fps being reported (collect-fps is using `requestAnimationFrame` under the hood). I tested this with https://gist.github.com/xaviervia/c17859ce49807f2cf8abfd14e2a699d5

Your comment gave me an idea though and I tested triggering the collection of FPS, then focusing another window, waiting for a couple of seconds and giving focus back to the windows collecting. Then the FPS count dropped significantly, to 13 FPS, which I’m guessing means that `requestAnimationFrame` doesn’t call its callback if the window is not focused. @pirelenito let’s sync about this.

Nevon:
You could maybe use the Page Visibility API to determine whether or not the benchmark is valid. Not supported in IE<10 though.

xaviervia:
Neither is `requestAnimationFrame` ;)

Nevon:
Yeah, but you can sort of polyfill `requestAnimationFrame`. Can't do that for the Page Visibility API as far as I know.

xaviervia:
Waiting on the result of https://github.com/pirelenito/collect-fps/pull/2

pirelenito:
There you go https://github.com/batata-frita/collect-fps/releases/tag/v1.1.0

xaviervia:
Now waiting for https://github.com/klarna/higher-order-components/pull/8

I’ll start adapting the API anyway since we do know that we want to move in that direction

Nevon:
All the dependencies of this has been merged now. Is this happening, or is there something other than the failed build that's blocking?

xaviervia:
@Nevon not sure about this one: https://github.com/klarna/higher-order-components/pull/10/files

But I guess. We can wait on that one until the renaming happens. For now we can merge this I guess.

xaviervia:
disable-animations-on-low-fps mov

no-radio-transition mov

xaviervia:
Hm, it wouldn't be a breaking change for the UI API though. Radio does not expose any of these externally. It would be breaking for Collapsible (which is not exactly stable and nobody is using it) but that one is going to be extracted into a separate package anyway (along with other animation components as we add them).

Unless I missed something this shouldn’t be a problem. Of course, for the higher-order-components this will be _super breaking_ but those are below v1 anyway and not directly consumed by anyone that I’m aware of.

Nevon:
I was thinking about the Collapsible, but if no one is using it, then we should be all good.

PR #274, opened 2017-01-20
Selector Input

cskeppstedt:
This is a new component. Its purpose is to:
- Look and behave much like an `<Input>`, but without the editable "input" field
- Show that a value can be "selected", and
- Show the "selected" value if there is one.

The selection itself is _not_ connected to the component. Here's a concept where the fullscreen dialog transitions to a `Selector.Options` component to do the selection:

selector concept

This has been pretty much a copy-paste from `<Input>`, which means that it has support for modes like `big`, `giant`, `icon` etc., it's stackable inside fieldsets, and so on. All variations haven't been confirmed or optimized, since it's not going to be needed right now. I left it in there anyway so it would be easy to fine-tune it when needed. It also supports dynamic/merchant/themeable styling, which is a requirement.

While it works, it's a bit crude, so I'll see if I can clean it up a bit. There's likely some stuff from Input still in there that is not needed.

### DIFF
--- a/Button/example.jsx
+++ b/Button/example.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Back, Close, Select, Hamburger, Options, Search } from '../IconButton'
+import { Back, Close, Hamburger, Options, Search, Select } from '../IconButton'
 import * as Block from '../Block'
 import * as Button from '../Button'
 // import * as Paragraph from '../Paragraph'
@@ -229,7 +229,7 @@ export default {
 
     {
       title: 'Iconic Buttons',
-      require: 'import { Back, Close, Hamburger, Options, Search } from \'@klarna/ui/IconButton\'',
+      require: 'import { Back, Close, Hamburger, Options, Search, Select } from \'@klarna/ui/IconButton\'',
       type: LIVE,
 
       examples: {
--- a/Button/example.jsx
+++ b/Button/example.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Back, Close, Hamburger, Options, Search } from '../IconButton'
+import { Back, Close, Select, Hamburger, Options, Search } from '../IconButton'
 import * as Block from '../Block'
 import * as Button from '../Button'
 // import * as Paragraph from '../Paragraph'
@@ -263,6 +263,10 @@ export default {
           </Block.Plain>,
 
           <Block.Plain key='1' style={{padding: '20px'}}>
+            <Select label='Select' />
+          </Block.Plain>,
+
+          <Block.Plain key='2' style={{padding: '20px'}}>
             <Close label='Close' color='gray' left />
           </Block.Plain>
         ],
--- a/Selector/Input/index.jsx
+++ b/Selector/Input/index.jsx
@@ -1,5 +1,6 @@
 import React, { PropTypes } from 'react'
 import classNamesBind from 'classnames/bind'
+import compose from '../../lib/compose'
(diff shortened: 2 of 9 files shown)


Selector/example.jsx
```diff
@@ -55,6 +57,46 @@ import UncontrolledSelector from '@klarna/ui/uncontrolled/Selector'`,
...
+    {
+      title: 'Selector Input',
+      require: `import Fieldset from '@klarna/ui/Fieldset'
+import * as Selector from '@klarna/ui/Selector'`,
+      type: LIVE,
+
+      examples: {
+        Variations: <Fieldset margins>
+          <Selector.Input
+            onClick={(e) => { console.info('it was clicked', e) }}
```
xaviervia: Changing this to `onClick={(e) => console.info('it was clicked', e)}` makes it render more pretty in the JSX output. I know it's misleading because `console.info` is not actually returning anything, this is just to make it more concise in the example.

Selector/Input/index.jsx
```diff
@@ -0,0 +1,224 @@
...
+import defaultStyles from './styles.scss'
+import * as fieldStates from '../../lib/features/fieldStates'
+import * as inlinedIcon from '../../lib/features/inlinedIcon'
+import * as stacking from '../../lib/features/stacking'
+import MouseflowExclude from '../../MouseflowExclude'
+import { Select } from '../../IconButton'
+import themeable from '../../decorators/themeable'
+import overridable from '../../decorators/overridable'
+
+const baseClass = 'input'
```
xaviervia: Let's do a simple rename of this class to `selector-input`. Some people are not using CSS modules, (most everyone is using them in production, but it should work even if we are not using them)

Button/example.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Back, Close, Hamburger, Options, Search } from '../IconButton'
+import { Back, Close, Select, Hamburger, Options, Search } from '../IconButton'
```
xaviervia: This import should be also in the example import statement

screen shot 2017-01-22 at 11 19 43

Selector/Input/styles.scss
```diff
@@ -0,0 +1,470 @@
...
+      content: '';
+      margin-top: -10px;
+      position: absolute;
+      right: 15px;
+      top: 50%;
+    }
+  }
+}
+
+%input__label {
```
xaviervia: Let's also rename this templates to `%select-input<whatever>`. They are repeated from `input`, but extracting them might be problematic, usually duplicating these works better.

cskeppstedt:
Thanks for the review @xaviervia! All valid points, I'll address them now.

Button/example.jsx
cskeppstedt: Aw yes, fixed in e769337

Selector/Input/index.jsx
cskeppstedt: Fixed in 0e14961

Selector/Input/styles.scss
cskeppstedt: Fixed in 0e14961

Selector/example.jsx
cskeppstedt: The wonders of code generation :) 75815e9

cskeppstedt:
I could say the same! How do I know _you_ are not a robot..?

xaviervia:
If a robot is better than a human at distinguishing robots from humans, what then?

PR #282, opened 2017-01-24
Switch to headless Electron as the Karma default browser

xaviervia:
The purpose of this is to remove PhantomJS, which is a source of issues in the pipeline, and use a headless Electron instance instead for running tests.

The implementation was done based on the guides:
- https://webuild.envato.com/blog/running-headless-javascript-testing-with-electron-on-any-ci-server/ 
- https://github.com/juliangruber/browser-run#headless-testing

Karma Electron: https://github.com/twolfson/karma-electron


xaviervia:
It seems that this runs smoothly in other machines, there is just some slowness when installing, but that's about it. I'll merge this and see if #121 can become ready.

PR #288, opened 2017-01-27
Calculate border-radii for all variations of stacked fields

cskeppstedt:
The existing implementation only concerned top-, bottom-, and square fields. This adds support for left- and right variations as well.

border-radii

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 npm-debug.log
+yarn-error.log
 .idea/
 *.woff
 *.eot
--- a/lib/features/stacking/index.js
+++ b/lib/features/stacking/index.js
@@ -2,9 +2,6 @@ import { PropTypes } from 'react'
 import combinations from '../../../lib/combinations'
 import fieldSizeFraction from '../../../propTypes/fieldSizeFraction'
 
-function borderRadii ({ top, bottom, left, right, square }, radius) {
-}
-
 export const position = {
   defaultProps: {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 npm-debug.log
+yarn-error.log
```
cskeppstedt: Almost committed the error log :)

lib/features/stacking/index.js
```diff
@@ -21,29 +21,20 @@ export const position = {
...
-        }
-        : square
-          ? {
-            borderRadius: '0px'
-          }
-          : {
-            borderRadius: radius
-          }
-  ),
+  getBorderRadii: ({ top, bottom, left, right, square }, radius) => {
```
cskeppstedt: lmk if you come up with a better/cleaner/more understandable way to deduce what border radii to return, I did a few iterations to arrive at this rule-ish approach.

xaviervia: This looks great. Except maybe coming up with some higher kinded polymorphism in the category of bifunctors, I can imagine a clean solution to this problem. And then it wouldn't be understandable anyway. This is about the best I can imagine.

cskeppstedt:
Wow

lib/features/stacking/index.js
cskeppstedt: My thinking exactly :)   👌

PR #290, opened 2017-02-02
Recalculate highlight on resize for Menu.Tab

xaviervia:
so-moloch-had-mercy-on-me mov

### DIFF
--- a/Menu/Tab/index.jsx
+++ b/Menu/Tab/index.jsx
@@ -74,12 +74,7 @@ export default React.createClass({
   componentDidMount (prevProps) {
     setTimeout(() => update(this))
 
-    window.addEventListener(
-      'resize',
-      () => window.requestAnimationFrame
-        ? window.requestAnimationFrame(() => update(this))
-        : update(this)
-    )
+    window.addEventListener('resize', this.resizeListener)
 
     if (
       this.props.focus &&
@@ -87,6 +82,16 @@ export default React.createClass({
     ) {
       this.refs[this.props.focus].focus()
     }
+  },
+
+  componentWillUnmount () {
+    window.removeEventListener('resize', this.resizeListener)
+  },
+
+  resizeListener () {
+    window.requestAnimationFrame
+      ? window.requestAnimationFrame(() => update(this))
+      : update(this)
   },
 
   render () {
--- a/Menu/Tab/index.jsx
+++ b/Menu/Tab/index.jsx
@@ -32,11 +32,7 @@ export default class Tab extends Component {
   constructor () {
     super()
 
-    this.resizeListener = debounce(() => {
-      window.requestAnimationFrame
-        ? window.requestAnimationFrame(() => update(this))
-        : update(this)
-    })
+    this.resizeListener = debounce(() update(this))
 
     this.state = {
       left: 0,
--- a/Menu/Tab/index.jsx
+++ b/Menu/Tab/index.jsx
@@ -3,6 +3,7 @@ import { Motion, spring } from 'react-motion'
 import classNamesBind from 'classnames/bind'
 import defaultStyles from './styles.scss'
 import getActiveElement from '../../lib/getActiveElement'
+import debounce from '../../lib/debounce'
 
 const baseClass = 'tab-menu'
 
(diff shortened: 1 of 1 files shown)


Menu/Tab/index.jsx
```diff
@@ -82,6 +84,16 @@ export default React.createClass({
     }
   },
 
+  componentWillUnmount () {
+    window.removeEventListener('resize', this.resizeListener)
+  },
+
+  resizeListener () {
+    window.requestAnimationFrame
+      ? window.requestAnimationFrame(() => update(this))
```
cskeppstedt: This would call `requestAnimationFrame` on every `resize` callback, so I assume it would queue potentially _lots_ of update calls?

xaviervia: Resize is actually not trigger that often. A quick try with Chrome reveals that trying hard to move it a lot I can get 50 resize events more or less. It would be even less on a less performant computer.

Do you think I we should debounce it? Doesn't the browser cancel previous updates if they get too delayed?

Nevon: That is correct. You have to do something like this instead (untested, but you get the idea):

```
const whenRendering = (fn) => {
  let scheduled = false
  
  return (...args) => {
    if (!scheduled) {
      requestAnimationFrame(() => {
        fn(...args))
        scheduled = false
      }
      scheduled = true
    }
  } 
}

// in `componentWillMount`
window.addEventListener('resize', whenRendering(() => update(this)))
```

xaviervia: Why do I feel like this helper is already implemented somewhere on the codebase. Couldn't find it, but maybe it was before?

Nevon: Yeah, I distinctly remember writing something like this before, but I don't remember where... Maybe somewhere in the instalments component? 😄

Nevon: About the resize, yes, you absolutely should delay execution until render. The reason you are seeing the resize handler invoked so few times is because that's the maximum your computer is able to handle. If you tried it on some shitty computer from 10 years ago running Safari 7, you'd only see it invoked like twice.

And no, the updates will not be cancelled. If the resize event has managed to happen, it will throw your callback onto the event loop and invoke it the next time the stack is empty. It's the same thing as if you do something like: `setInterval(() => setTimeout(() => console.log('lol'), 0), 0)`

Nevon: Found it! https://github.com/klarna/ui/blob/0ff20fbcce4b6013723f73d9135740ce2ab194e0/lib/debounce.js

xaviervia: Awesome. I'll wrap it and push again

xaviervia: Done. My quick test showed it working. Whether or not this is making a difference I am not sure how to test (that is, without investing a lot of time ;) )

Menu/Tab/index.jsx
```diff
@@ -82,6 +85,16 @@ export default React.createClass({
     }
   },
 
+  componentWillUnmount () {
+    window.removeEventListener('resize', debounce(this.resizeListener))
```
Nevon: This won't work, since the reference to the callback won't be the same. In the constructor, you'll have to create an instance variable for the debounced listener and use that here and in `componentDidMount`.

Nevon: Actually, now that I look at the diff, you're not using the debounced listener anywhere except here :P

xaviervia: OMG. I should go and sleep xD

Menu/Tab/index.jsx
```diff
@@ -31,6 +31,13 @@ const update = (component) => {
 export default class Tab extends Component {
   constructor () {
     super()
+
+    this.resizeListener = debounce(() => {
+      window.requestAnimationFrame
```
Nevon: You don't need to call `requestAnimationFrame` inside this. It's already done inside the `debounce`. What's going to happen now is that when the resize event happens, it schedules a callback for the next render. Once that render callback is called, it will schedule a callback for the next render. Finally in that render your update will happen.

Just do:

```
this.resizeListener = debounce(() => update(this))
```

xaviervia: 😊

PR #293, opened 2017-02-03
Fix Collapsible allowing content to overflow

pirelenito:
The current implementation leaves all Collapsibles as `overflow: visible` on mount. However it only manifest itself as an issue in IE11.

- Move around the documentation explaining the reasoning
- Add an extra rule that keeps the overflow hidden while collapsed

### DIFF
--- a/Collapsible/index.jsx
+++ b/Collapsible/index.jsx
@@ -32,14 +32,9 @@ export default class Collapsible extends Component {
       }}>
         {({height, opacity}) => <div
           style={{
-            // once it is fully expanded, we set the heigh to auto
-            // and let the browser take care of the size
             height: getHeight(collapsed, height, this.state.height),
             opacity,
-            // Overflow rule to enable content to overflow outside the collapsible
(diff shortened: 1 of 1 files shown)


Collapsible/index.jsx
```diff
@@ -48,6 +43,20 @@ export default class Collapsible extends Component {
   }
 }
 
+/**
+ * Overflow rule to enable content to overflow outside the collapsible
+ * once the animation is close to be complete (the last few pixels take a while
+ * to be expanded). '10' is a magic number 🎩
+ */
+const shouldOverflow = (collapsed, animatedHeight, actualHeight) => {
+  const notYetAlmostExpanded = actualHeight - animatedHeight > 10
```
Nevon: This is a tiny little comment, and I think either way is okay, but for me personally I think a name like `notYetAlmostExpanded` is one negation too many. To me, it would be more clear if you reversed the comparison and then did `(collapsed || !almostExpanded)`.

PR #295, opened 2017-02-07
Smoother animations on Radio components

pirelenito:
Introduces a new component `RadioMark` that implements the radio circle as an SVG component.

It also fixes an issue on Android stock where the Radio component would not render the checked state correctly.

radio-animation

Me and @xaviervia verified the new transition with @hannah-ahl.

**This PR is based on the performance PR that needs to be merged first. We will be doing some regression on this PR and then merge both on success.**


RadioMark/index.js
```diff
@@ -0,0 +1,64 @@
...
+import themeable from '../decorators/themeable'
+
+const grayLines = '#cbcbcd'
+const klarnaBlue = '#0074c8'
+const white = 'white'
+
+function RadioMark ({ customize, checked, disabled, onClick, lowFPS }) {
+  const backgroundColor = customize ? customize.backgroundColor : klarnaBlue
+  const bulletColor = customize ? customize.bulletColor : white
+  const scale = checked ? 1 : 3
```
Nevon: I think these magic numbers should have names, like `SCALE_FACTOR_[CHECKED|UNCHECKED]` or something.

RadioMark/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <svg width='14' height='14' onClick={onClick}>
+      <Motion style={{ scale: lowFPS ? scale : spring(scale) }}>
+        {({ scale }) => (
+          <g>
+            <circle cx='7' cy='7' r='7' fill={grayLines} />
+            <circle cx='7' cy='7' r='7' fill={disabled ? grayLines : backgroundColor} style={{ opacity: 1 - (scale - 1) / 2 }} />
+            {/* tranformOrigin using matrices: http://stackoverflow.com/questions/6711610/how-to-set-transform-origin-in-svg#6714140 */}
+            <g transform={`matrix(${scale}, 0, 0, ${scale}, ${7 - scale * 7}, ${7 - scale * 7})`}>
```
Nevon: Got a lot of magic numbers here. What's the significance of 7? Is it the circle diameter? Then it should probably be extracted and referenced from the circle as well.

Showroom/examples/components.js
```diff
@@ -19,6 +19,7 @@ export MouseflowExclude from '../../MouseflowExclude/example'
 export Preview from '../../Preview/example'
 export ProgressBar from '../../ProgressBar/example'
 export Radio from '../../Radio/example'
+export RadioMark from '../../RadioMark/example'
```
Nevon: Is the RadioMark valid in any other context than within a `Radio`? If not, I don't think it should be a top level component.

PR #296, opened 2017-02-10
Partially collapsed Radio selector

xaviervia:
Notes to the reviewer:

- This has been tested in IE 9, IE 10, iOS simulator, Safari Firefox and Chrome desktop. It would be ideal to give it a quick run in Android Stock browser, but other than that looks fairly promising.
- There are [two](https://github.com/klarna/ui/pull/296/files#diff-b5545cd74975f65dbd1829cc6ded71fc) [new](https://github.com/klarna/ui/pull/296/files#diff-a0b50c48a2153dd6bfce6f5b43e1d7eb) higher-order components added to deal with inline styles for the new ExpandLabel component (Radio only, which is why it doesn’t have a place in the showroom). These two higher-order components **should be moved to [@klarna/higher-order-components](https://github.com/klarna/higher-order-components)**—the reason they are not right now is that it can be done in the next release.
- The extra white layer in the animation that makes the ExpandLabel box disappear is there on purpose. If you have questions about the animation you can also ping @koorochkin 

chrome mov

### DIFF
--- a/Radio/example.jsx
+++ b/Radio/example.jsx
@@ -1,4 +1,6 @@
 import React from 'react'
+import compose from 'ramda/src/compose'
+import {withState, withHandlers} from 'recompose'
 import Radio from '../Radio'
 import * as Checklist from '../Checklist'
 import * as Paragraph from '../Paragraph'
@@ -72,6 +74,16 @@ const optionsWithContent = [
   }
 ]
 
+const StateFullyCollapsed = compose(
+  withState('fullyExpanded', 'onExpand', false),
+  withHandlers({
+    onExpand: props => () => {
+      console.log('this is being called')
+      props.onExpand(!props.fullyExpanded)
+    }
+  })
+)(Radio)
+
 export default {
   title: 'Radio',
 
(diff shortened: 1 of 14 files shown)


xaviervia:
What it should actually happen is that the payment selector height is animated based on the `fullyExpanded` flag. Splitting into two structures won’t work. We probably need something like the `Collapsible` but that accepts custom heights, either parametrized or inspecting its children.

Radio/example.jsx
```diff
@@ -99,6 +105,13 @@ export default {
         value='ipsum'
       />,
 
+      'Partially collapsed & controlled': <Radio
+        options={optionsWithContent}
+        expandLabel='Show me all the options'
+        collapsedOptions={2}
```
Nevon: This feels backwards. Wouldn't it make more sense to have this be `visibleOptions` or something like that? If I have a dynamic list of options, I would rather say that I want three of them to be visible than `options.length  - Math.max(options.length - 3, 0)` collapsed.

Radio/index.jsx
```diff
@@ -225,14 +216,200 @@ const Radio = React.createClass({
             </div>
           ]
         })}
+
+        {collapsedOptions && <Collapsible
+          onStartFPSCollection={onStartFPSCollection}
+          onEndFPSCollection={onEndFPSCollection}
+          lowFPS={lowFPS}
+          minimumHeight={49}
```
Nevon: What is `minimumHeight` and what is this magic number?

Radio/index.jsx
```diff
@@ -225,14 +216,200 @@ const Radio = React.createClass({
...
+          <div>
+            {optionLists.collapsed.map((option) => {
+              const {
+                key,
+                label,
+                description,
+                disabled,
+                aside,
+                content,
+                leftPad,
```
Nevon: 😂

Radio/index.jsx
```diff
@@ -225,14 +216,200 @@ const Radio = React.createClass({
...
+                  id={ids.label}
+                  {...restOfProps}>
+                  <label
+                    htmlFor={`${name}-${key}`}
+                    className={classNames(classes.optionHeader)}
+                    id={ids.header}>
+                    <div
+                      className={classNames(classes.optionHeaderInner)}
+                      id={ids.headerInner}>
+                      {!singleOption && <div className={classNames(classes.optionLeft, classes.optionLeftmost)} id={ids.left}>
```
Nevon: Does the `Radio` have a horizontal mode, or why is this called `*LeftMost`?

Nevon: And same for the `RightMost`

Radio/index.jsx
```diff
@@ -225,14 +216,200 @@ const Radio = React.createClass({
...
+                </div>
+              ]
+            })}
+          </div>
+        </Collapsible>}
+
+        {expandLabel && <Motion
+          style={{
+            opacity: spring(isExpanded ? 0 : 1),
+            height: spring(isExpanded ? 0 : 49, {stiffness: 40, damping: 15})
```
Nevon: Is this `49` same as the other magic `49`?

settings/fontFamilies.js
```diff
@@ -0,0 +1,3 @@
+export const BASE = '\'Open Sans\', \'Helvetica Neue\', \'Helvetica\', \'Arial\', sans-serif'
```
Nevon: If you use template strings instead, you don't have to bother with escaping those quotes.

Radio/example.jsx
xaviervia: Sure. I can invert that API. Sounds reasonable, I'm not sure why I did this in the first place.

Radio/index.jsx
xaviervia: Hm. So, Collapsible now supports a starting height. The magic number is the height of the ExpandLabel, which unfortunately would be rather hard to capture (but I should try, yes, I know. I will)

Radio/index.jsx
xaviervia: I think the pun was intended. It's probably @pirelenito  ;)

Radio/index.jsx
Nevon: As a start, you could extract it to a named constant so that you at least can know what it represent, and make sure that it's also used in that other place.

Radio/index.jsx
xaviervia: This diff is green because I copy pasted. The decision was taken before. I know though it's a good opportunity for refactoring this "option" out, but I think that should also imply updating the styling strategy to inline (because in the current one the CSS is so tangled that extracting is hard). 

I don’t know what the deadline status is, maybe it's better to do it in a second pass.

settings/fontFamilies.js
xaviervia: If I do that the linter tells me that it is disappointed with my style :(

Radio/example.jsx
xaviervia: This would be a breaking change if I don't do it now, so I will do it now.

xaviervia: Sorry, what I mean to imply is that some other changes I will do in a PR targeting this PR, so that we can safely merge this one and release of deadline so demands it.

Nevon: Deadline Driven Development. I've heard on the internets that DDD is a good thing, so 👍

settings/fontFamilies.js
xaviervia: Never mind. The last update of standardjs is now aware of this use case. I will fix 😄

PR #301, opened 2017-02-13
Remove selectedBorderColor from hover dropdown

beveryday:
We do not want this customization in place yet per Vladimir Kurochkin.


xaviervia:
Let's fix this in the consuming product :) I’ll close for now

PR #304, opened 2017-02-15
Progress bar animation

ChristerGustavsson:
progress-bar-animation


Nevon:
The example is a bit confusing, since the `TimedExample` component is not a component that we expose. I don't remember how the documentation is rendered (@xaviervia probably knows) , but it would be good to show that in the example.

ProgressBar/example.jsx
```diff
@@ -2,6 +2,28 @@ import React from 'react'
...
+class TimedExample extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      percentage: 0
+    }
+  }
+
+  componentDidMount () {
+    setInterval(() => {
```
Nevon: Unbind this in `componentWillUnmount`. No need to trick people into adding memory leaks.

JoakimLofgren:
@Nevon @xaviervia Oh right, yeah, you could overwrite the JSX string shown in the example right?
I guess we should do that and just display with value + partial props?

xaviervia:
showcase mov

@JoakimLofgren I decided to try another approach, since this is not so much of a feature, rather a demonstration. I grabbed a layout that @hannah-ahl created for the Showroom that I never had the chance to implement. Wdyt?

Nevon:
... Can you make an equalizer?

xaviervia:
@Nevon you mean representing the waves of your mic input as progress bars, or actually allowing you to use the bars to transform the sound? :D

Nevon:
¿Porque no los dos?

xaviervia:
@Nevon wdyt of the updates?

PR #305, opened 2017-02-16
Fix code examples for 'real' focus

xaviervia:
The `focus` prop was missing the `real` value. Thanks @MikaelAmborn


MikaelAmborn:
If you leave the `focus='real'` field by tabing out of it or clicking on another field it will regain focus if you move your mouse cursor over the field.

xaviervia:
@MikaelAmborn the behavior is correct, because the Field is fully controlled, so it should set focus to itself automatically on any interaction. However, I can see how this is not a great experience in the Showroom 😔 . I wonder what the alternative is.

PR #306, opened 2017-02-16
Fix broken IconButton on android

cskeppstedt:
Improves the IconButton on Android 4.x devices. Apparently, having `padding` on an `<svg>` causes a lot of problems on older Android devices. I found this by spending much of my sanity in the Android emulator, tweaking the various css properties for `IconButton`.

As a small bonus, the `:active` hint (the circle with the background color) now also works on IE 10/11, by adding `pointer-events` to the svg to let the click event fall-through to the divs behind it. However it still only works if you click the actual icon, not the text. I assume this has to do with how the `:active` pseudo selector is implemented. But, I think this is a minor improvement so I kept it.

android4iconbutton

### DIFF
--- a/IconButton/Back/index.jsx
+++ b/IconButton/Back/index.jsx
@@ -7,6 +7,7 @@ import compose from 'ramda/src/compose'
 import {overridable, themeable} from '@klarna/higher-order-components'
 
 const classes = {
+  bgWrapper: 'bg-wrapper',
   iconButton: 'icon-button',
   fill: 'illustration__fill',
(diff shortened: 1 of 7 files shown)


cskeppstedt:
I tested the showroom in a range of browsers, Android 4.3, IE9-11, mobile safari, etc. It looks good. I'll do device-testing inside KC to see that nothing broke there before merging.

Also, I wanted to try to do this without adding one more `div` but couldn't find a way to do that.

PR #307, opened 2017-02-17
Default percentage to 0

matthiasfeist:
First mount of component had issues with percentage stuck at 100


xaviervia:
@ilonashub @benzaita @benklarna @omriyariv @ohadco this will affect the way ProgressBars render in MyKlarna, since they will always animate from 0 to the initial value. Here is a very ugly gif of how that looks like. Wdyt?

defaultanimation mov

matthiasfeist:
So maybe we can make the start value into a property? Does that solve our needs?

PR #310, opened 2017-02-18
Remove all peer dependencies

### LINKED ISSUE #309: Peer dependencies
How come we are listing react, react-motion and ramda as peer dependencies? The purpose of peer dependencies is to say that you depend on a package even though you never actually `require` it. The canonical use-case is for plugins, so that if you for example build a webpack plugin, you would list webpack as a peer dependency to be able to express that you are compatible with a certain version of webpack, even though you never `require` webpack in your code.

The UI components certainly `require` all of react, react-motion and ramda, so they should be listed as dependencies (which they already are, which makes me even more confused about why they are also listed as peer dependencies).

Nevon:
Fixes #309

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,11 +81,6 @@
     "yargs": "^3.31.0",
     "zen-router": "^1.0.0"
   },
-  "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0",
-    "react-motion": "^0.4.2",
-    "ramda": "^0.21.0"
-  },
   "standard": {
     "globals": [
       "describe",


pirelenito:
If you don't install them, the UI won't build. So it is better to have the package manager warning us.

Nevon:
Yeah, my point was that that's not what peerDependencies are for. It should just be a regular dependency. See #309.

pirelenito:
If they were just a dependency, we can have multiple versions of the same library in the same bundle. Peer depedencies is also a way of enforcing an specific version and making sure it is only installed once.

I'm not sure it is incorrect.

PR #317, opened 2017-02-22
Make the default BoxSelector overridable

npejo:
The default BoxSelector component switches between the Horizontal and the Vertical component depending on the screen size. But since it was not overridable, even the nested components were overridable, the custom designs are not propagated properly.

I tested, it looks that only the `overridable` high-order component is needed, it works fine without `themeable` or `uncontrolled`.

### DIFF
--- a/BoxSelector/index.jsx
+++ b/BoxSelector/index.jsx
@@ -60,4 +60,4 @@ BoxSelector.propTypes = {
   layout: PropTypes.oneOf(['auto', 'horizontal', 'vertical'])
 }
 
-export default overridable(defaultStyles, BoxSelector)
+export default overridable(defaultStyles)(BoxSelector)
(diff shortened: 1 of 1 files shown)


BoxSelector/index.jsx
```diff
@@ -55,4 +61,6 @@ BoxSelector.propTypes = {
   layout: PropTypes.oneOf(['auto', 'horizontal', 'vertical'])
 }
 
-export default BoxSelector
```
joaomilho: The way I understand it `compose(x(A))(B)` is the same as `x(A, B)`. So unless you need more functions `compose` is redundant here.

npejo: Thanks @joaomilho , you are right. `compose` end up as leftover of the initial solution :)

npejo: fixed in 423b073a08743a3f

PR #322, opened 2017-02-23
Add optional 'title' property to Checklist component

marcioos:
screen shot 2017-02-23 at 14 46 19
```javascript
<Checklist.Main title='TITLE'>
  <Checklist.Item>Item1</Checklist.Item>
  <Checklist.Item>Item2</Checklist.Item>
  <Checklist.Item>Item3</Checklist.Item>
</Checklist.Main>
```

### DIFF
--- a/Checklist/index.jsx
+++ b/Checklist/index.jsx
@@ -2,6 +2,7 @@ import React, { PropTypes } from 'react'
 import classNamesBind from 'classnames/bind'
 import defaultStyles from './styles.scss'
 import childrenPropType from '../propTypes/children'
+import * as Paragraph from '../Paragraph'
 
 import compose from 'ramda/src/compose'
 import {overridable, themeable} from '@klarna/higher-order-components'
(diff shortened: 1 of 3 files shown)


Nevon:
It feels like this should be a child component, not a prop. Something like `Checklist.Title`.

Checklist/index.jsx
```diff
@@ -29,10 +31,21 @@ function ChecklistMain ({ chromeless, className, children, customize, style, sty
...
+    {title && renderTitle(title, classNames(classes.title))}
     {children}
   </ul>
 }
 
+function renderTitle (title, className) {
+  return (
+    <li>
+      <Paragraph.Primary className={className}>
+        <strong>{title}</strong>
```
Nevon: Get rid of the `strong` and apply the style change on the class instead.

xaviervia: We are rejecting this one now since the `Block.Installments` component already behaves like it was intended.

marcioos:
@Nevon As @xaviervia mentioned, we can achieve the same behavior using `Block.Installments` with a chromeless `Checklist` inside. So no need for this change.

Issue #321, opened 2017-02-23
Introduce `link` variation for buttons and better Landing API


xaviervia:
## For the Button

If an `href` prop is specified, let’s render an `<a>` instead of a `<button>`

## For Landing

Let’s change the API to:

```jsx
<Landing
  texts={{ ... }}
  actions={{
    accept: { text: '', href: '', ...acceptProps },
    cancel: { text: '', ...cancelProps }
  }}
  onAccept={() => {}}
  onCancel={() => {}}
/>
```

PR #323, opened 2017-02-24
Support link in button

xaviervia:
`Button` components (`Primary`, `Secondary` and `Tertiary`) now support the `href` and `target` props that will cause them to render as `<a>`.

screen shot 2017-02-27 at 11 03 46

Fixes https://github.com/klarna/ui/issues/321


xaviervia:
As @schold suggested, let's add `role="button"` and `role="link"` depending on the semantics.

Issue #62, opened 2016-09-25
Fix z-index hell


joaomilho:
<img width="737" alt="screen shot 2016-09-25 at 3 19 28 pm" src="https://cloud.githubusercontent.com/assets/308196/18815264/7fee1c40-8333-11e6-86ec-9ec1f6759675.png">

We shouldn't be setting arbitrary z-index numbers. Ideally we should use scss's array feature to define were things stand, like this:

```scss
$indexes = '.button' '.field' ...
@mixin z-index($element) {
  z-index: index($indexes, #{$element});
}
```

joaomilho:
Ping @ilonashub @xaviervia

xaviervia:
I'm not sure I understand. I'd do this with a map structure and `map-get`. Unless there is some magic that I'm missing.

xaviervia:
First analysis attempt of the z-indexes we use:

button:
  v = 1
  to be above the black cover

dialog:
  v = -1
  to be hidden
  v = 100
  to ensure it has time to transition before showing
  so hackish

dropdown:
  v = 10
  hover, to be above neighboring
  v = 20
  focus: to be above neighboring + hover
  v = 5
  error, warning: to be above neighboring plain ones
  v = 2
  select tag: to be clickable/tappable

field:
  == dropdown, minus v = 2
  v = 10
  ie9 background image, ensure field is interactable and not label (no pointer-events prop)

input:
  == field
  v = 1
  ie9. reason unknown

box-selector:
  v = 5
  cell highlight should be above elements

menu-segmented:
  v = 5
  selected: be above the plain ones

PR #325, opened 2017-02-24
Full switch to inline styles

xaviervia:
After the fixes in Radio and many others that are moving towards inline styles, we should make a concerted effort of moving everything else to inline styles. This PR is meant for that.


xaviervia:
- [ ] Follow up https://github.com/klarna/ui/issues/62#issuecomment-261624816

xaviervia:
This will be done by rebuilding from scratch

PR #336, opened 2017-03-01
Introduces template CenteredDialog

schold:
In addition to what's offered in the chromes/Centered template, I wanted a dialog with a looser interface regarding passing buttons. 

This also utilizes html2KlarnaUI for summary text, which will significantly improve the handling of localization strings in React.


xaviervia:
@schold please do :) . You can add them in here https://github.com/klarna/ui/blob/master/templates/examples/Informative.jsx 

After that I think we are ready to merge.

PR #347, opened 2017-03-10
Add responsive component

xaviervia:
TBD


xaviervia:
In the end we are using https://github.com/ctrlplusb/react-component-queries which is a more comprehensive solution for measuring the container size and reacting on it. I'm rejecting this.

PR #352, opened 2017-03-15
Slightly better fail for long button labels in responsive context

schold:
Not the most graceful solution. Don't know how much I'll break. 

Gives an alternative approach to how badly buttons fail in a responsive context for multi-word labels.

We could perhaps build on this?


xaviervia:
Sic transit gloria mundi

PR #357, opened 2017-03-23
Adjusts customize behavior for tertiary button 

schold:
So that it mimics the behavior for when customizations has not been applied.


xaviervia:
Screenshot? Code looks good

schold:
@xaviervia Here it is.
**Normal**
screen shot 2017-03-29 at 12 24 01
**Hovered**
screen shot 2017-03-29 at 12 23 40
**Clicked**
screen shot 2017-03-29 at 12 25 16

PR #362, opened 2017-04-05
Add visual regression tests

deepsweet:
## Components

* [ ] Alert
* [ ] Block
* [ ] BoxSelector
* [ ] Button
* [ ] Checklist
* [ ] Collapsible
* [ ] ContextMenu
* [ ] Dialog
* [ ] Dropdown
* [ ] Field
* [ ] Fieldset
* [ ] html2KlarnaUI
* [ ] Icons
* [ ] Input
* [ ] Label
* [ ] List
* [ ] Loader
* [ ] Menu
* [ ] MouseflowExclude
* [ ] Preview
* [ ] ProgressBar
* [ ] Radio
* [ ] RadioMark
* [ ] SavedInputLabel
* [ ] Selector
* [ ] Strong
* [ ] Switch
* [ ] Text
* [ ] Theme
* [ ] Tooltip

## Templates

* [ ] Informative
* [ ] Prompts
* [ ] Selection

## Compositions

* [ ] UncontrolledForm
* [ ] UniqueIds


xaviervia:
What is the status of this? How should we follow up?

deepsweet:
The conclusion: it's a bit complicated :)

1. Gemini has a limited API and it's difficult to extend it. We have switched to [Nightwatch](https://github.com/nightwatchjs/nightwatch) + custom diffing implementation using [pixelmatch](https://github.com/mapbox/pixelmatch)
2. [Git LFS](https://git-lfs.github.com/) should be used to store images, otherwise it will blow up the repo size drastically (as a binary format images can't be diffed across commits)
3. Geckodriver for Selenium has a lot of cons and quirks (especially with page interactions like clicks, mouse move, etc), we still can't get it work as stable and smooth as Chrome

I'm unable to finish this PR in the near future, it requires a lot more work and investigation than it was expected. Nightwatch is a better direction to go for now.

deepsweet:
Also, if only Chrome is enough then I'd suggest to try something like [chromy](https://github.com/OnetapInc/chromy) +  [pixelmatch](https://github.com/mapbox/pixelmatch) to avoid all this Selenium world as much as you can.

xaviervia:
🔥  ☄️💥

PR #364, opened 2017-04-05
Add Button responsive to width

beveryday:


xaviervia:
Thanks!

I added a small example to try it out. In general it looks good, but when made very narrow it becomes vertical for some reason. We can combine the `%` width in the wide version with some min-width to ensure it never goes below a certain width. This might also not be a problem for your use case.

screen shot 2017-04-06 at 10 41 56

screen shot 2017-04-06 at 10 42 01

beveryday:
I added a min-width of 154px. This is the smallest it gets in checkout. If someone wants to use the responsive class and needs it smaller I think they can change it without much harm to us.

PR #365, opened 2017-04-06
Adds a simple bordered block

schold:


Block/Bordered/styles.scss
```diff
@@ -0,0 +1,13 @@
+@import '../../css/settings';
+@import '../../css/mixins/index';
+
+.block--bordered {
+  background: map-get($colors, white);
```
xaviervia: Let's add `box-sizing: border-box` so the borders don't make it overflow outside the grid

Block/example.jsx
```diff
@@ -35,7 +35,12 @@ export default {
             Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris at aliquet leo.
           </Paragraph>
         </Block.Installments.Content>
-      </Block.Installments.Main>
+      </Block.Installments.Main>,
+
+      'Bordered': <Block.Bordered>
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris at aliquet leo.
```
xaviervia: Let's use a real `Paragraph.Secondary` as content in the example, so that the font looks alike in the Showroom

Block/Bordered/index.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { PropTypes } from 'react'
+import compose from 'ramda/src/compose'
+import classNamesBind from 'classnames/bind'
+import childrenPropType from '../../propTypes/children'
+import defaultStyles from './styles.scss'
+import {
+  overridable,
+  themeable,
+  uniqueName
```
xaviervia: Do we need this one?

Block/Bordered/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+Bordered.propTypes = {
+  children: childrenPropType,
+  className: PropTypes.string,
+  id: PropTypes.string,
+  styles: PropTypes.object
+}
+
+export default compose(
+  themeable((customizations, {customize}) => ({
```
xaviervia: It would be nice to have an example of a customized `Block.Bordered` with squared borders in the Showroom :)

Block/Bordered/styles.scss
schold: Ah, agreed.

Block/example.jsx
schold: Sure

Block/Bordered/index.jsx
schold: No. Removed it from the implementation, so should not be part of the imports. Will fix.

Block/Bordered/index.jsx
schold: Hehe, I actually did one. But never committed it.

Block/Bordered/styles.scss
```diff
@@ -0,0 +1,14 @@
...
+  border: ($grid * .2) solid $border-color;
+  border-radius: $border-radius;
+  box-sizing: border-box;
```
xaviervia: It’s still adding up to one more than the grid unfortunately. Let’s set one pixel less in the top padding:

`paddingTop: ($grid * 2.8)`

I also noticed that the margins on top and on bottom, when added to the default text paddings, make the vertical spacings look bigger than the horizontal ones. What if we make the top and bottom paddings one grid less?

```css
paddingBottom: ($grid * 2);
paddingLeft: ($grid * 3);
paddingRight: ($grid * 3);
paddingTop: ($grid * 1.8);
```

Block/example.jsx
```diff
@@ -31,11 +31,23 @@ export default {
...
-      </Block.Installments.Main>
+      </Block.Installments.Main>,
+
+      'Bordered': <Block.Bordered>
+        <Paragraph.Secondary>
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris at aliquet leo.
+        </Paragraph.Secondary>
+      </Block.Bordered>,
+
+      'Custom-bordered': <Block.Bordered style={{borderRadius: 0}}>
```
xaviervia: Since we can set this via `style`, we don't need the `customize` prop, right?

schold: Well, don't we? We need to handle the `customize` prop for when we are wrapped by the `Theme` component?

PR #368, opened 2017-04-13
Move react-hot-loader to dependencies

beveryday:
Our tests are breaking because of missing plugin.


xaviervia:
This will not mean that the Hot loader will be added to the final bundle, right?

xaviervia:
Closing this for now, let's fix it in the consuming project.

PR #369, opened 2017-04-17
Add field link for field component

beveryday:


Field/index.jsx
```diff
@@ -26,7 +27,8 @@ const classes = {
   iconInput: `${baseClass}--icon__input`,
   iconLabel: `${baseClass}--icon__label`,
   input: `${baseClass}__input`,
-  label: `${baseClass}__label`
+  label: `${baseClass}__label`,
+  skipLink: `${baseClass}--skip__link`
```
Deschtex: Is this class being used?

Field/index.jsx
```diff
@@ -225,7 +236,8 @@ const Field = React.createClass({
     const ids = id
       ? {
         input: `${id}__input`,
-        label: `${id}__label`
+        label: `${id}__label`,
+        link: `${id}__${fieldLink}__fieldLink`
```
Deschtex: I believe it should be written as `field-link` instead of `fieldLink`. WDYT?

FieldLink/index.jsx
```diff
@@ -0,0 +1,35 @@
...
+import classNamesBind from 'classnames/bind'
+import defaultStyles from './styles.scss'
+
+const baseClass = 'field-link'
+
+class FieldLink extends React.Component {
+  propTypes: {
+    id: PropTypes.string,
+    label: PropTypes.string.isRequired,
+    onFieldLinkClick: PropTypes.func
```
deepsweet: `isRequired`?

Field/index.jsx
beveryday: Sounds good to me. I wasn't sure

FieldLink/index.jsx
beveryday: Ya I think we can remove that

PR #375, opened 2017-04-26
Add field tooltips

beveryday:


Field/example.jsx
```diff
@@ -118,6 +118,20 @@ export default {
...
+        'With Field Tooltip': {
+          live: <Field
+            arrow='bottom-right'
+            disabled
+            fieldTooltip='We need your phone number for stuff'
+            label='Phone Number'
+          />,
+          code:
+`<Field
+  arrow='bottom-right'
```
beveryday: Talked through this in our sync.

Field/example.jsx
```diff
@@ -118,6 +118,20 @@ export default {
...
+  fieldTooltip='We need your phone number for stuff'
```
beveryday: Good Call

PR #376, opened 2017-04-27
Add key identifier for Radio/Option

WebCloud:
React conciliation script is mistakenly applying styles for options across different states for Radio/Option elements.

This patch will prevent it from happening.


skogby:
👍  :shipit:

PR #377, opened 2017-05-01
Remove direct export of imported dependencies

npejo:
Avoid using of stage-0 JS features

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,7 @@
   "presets": [
     "es2015",
     "react",
-    "stage-0"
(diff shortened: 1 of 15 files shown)


pirelenito:
We should probably also change the `.babelrc` and `package.json` to remove references to `stage-0`.

xaviervia:
Can I have some context?

xaviervia:
Pushed the change of stage-0 to stage-3 in both package.json and yarn.lock. Afaic this is ready to merge now :)

PR #381, opened 2017-05-15
Use extracted PropTypes

xaviervia:
Since the PropTypes have been extracted out of React, let's extract them here before they are deprecated.


DonGideon:
1p1ztq

PR #384, opened 2017-05-19
Add peer dependencies to externals

stipsan:
This is to avoid bundling a separate copy of React in this build, which
causes problems like this:
https://facebook.github.io/react/warnings/refs-must-have-owner.html

With this change it is possible to load kui without changing the webpack.config.js of projects that wish to use this library, and avoid having to add dependencies like `node-sass` (which has its own host of problems with having to run `npm rebuild node-sass` time and again).

Wonderful UI library you are open sourcing btw 🎉

### DIFF
--- a/webpack.config.build.js
+++ b/webpack.config.build.js
@@ -15,7 +15,14 @@ module.exports = {
     filename: '[name]-built.js',
     library: 'kui',
     libraryTarget: 'umd',
-    umdNamedDefine: true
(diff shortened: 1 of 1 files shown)


xaviervia:
Thanks! Travis seems to be failing for unrelated reasons, so I'll merge this and bump.

stipsan:
Thanks @xaviervia 😄

PR #385, opened 2017-05-23
Update to new React standards even from dependencies

xaviervia:
tl;dr React 16 will deprecate `React.createClass` and `React.PropTypes` so we need to ensure that nothing we depend on will break. We still need to find out specifically what needs to be fixed. 

---

Pending updates are:

- [x] `React.createClass` is still being used throughout the library. This should be changed.
- [ ] `React.createClass` is being used by some dependencies. Those dependencies have to be identified and updated or replaced.
- [ ] `React.PropTypes` (instead of standalone `PropTypes`) is still being used by several dependencies. Also need to be identified and updated or replaced.

### DIFF
--- a/BoxSelector/Horizontal/index.jsx
+++ b/BoxSelector/Horizontal/index.jsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {Component} from 'react'
 import PropTypes from 'prop-types'
 import classNamesBind from 'classnames/bind'
 import {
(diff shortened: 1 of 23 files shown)


xaviervia:
Regarding work on `react-maskedinput`, which seems to not be maintained anymore, we need to evaluate options. Currently testing https://text-mask.github.io/text-mask/ in several browsers:

- [x] Chrome (MacOS)
- [ ] Firefox (MacOS)
- [x] Safari (MacOS)
- [ ] Chrome (Windows)
- [ ] IE (Windows)
- [ ] Chrome (Android)
- [ ] Safari (iOS)

We know that it will not work for Android Stock, but that is also true for `react-maskedinput`.

xaviervia:
For `text-mask` 

- [ ] Find out if the React implementation is up to date with standards
- [ ] Try using it instead of `react-maskedinput` now

xaviervia:
Trying text-mask was in general successful, but there is an issue where the `ref` passed to the component will not be passed to `<input>` (https://github.com/text-mask/text-mask/blob/master/react/src/reactTextMask.js#L42) . This means we need to either PR to the `react-text-mask` project or work around it by setting the `ref` on the wrapper div and accessing the `input` with `this.wrapperDiv.querySelector('input')`. 

I say we do the second option and add an issue in the text-mask repo to have them aware of it.

skogby:
In checkout we currently use a inlined slightly modded version of: https://github.com/sanniassin/react-input-mask
The reason why we use it is that when focusing the input, it switches to show the mask directly (you don't have to type to see it). I personally think it works better with our overlay labels/placeholders and that it is a more clear way of showing what input is expected. Pleas come by and discuss for an example or more input! :)

xaviervia:
@skogby excellent point. Actually, normalizing the behavior of this type of fields is important to make the experience consistent. Let's agree on one approach and stick with it across the board.

Kadrei:
we sadly need more functionality than what react-input-mask brings. 
text-mask has the right amount of mask customizability. As far as I can see text-mask brings an option to always show guides: https://github.com/text-mask/text-mask/blob/master/componentDocumentation.md#guide

The difference is, that it doesn't only show on focus.

xaviervia:
I think we will face the need to update across the board either way we choose. In general though we should try to also align in the behavior of the field. It might not be possible, of course. It seems text-mask does not provide the functionality of showing the mask on focus, so it doesn't seem that would fit that use case, but on the other hand if we need the extra features of text-mask, we can't go with react-input-mask either.

stipsan:
Hey guys, we're trying to upgrade to React 16 and this library is blocking us :)

What I will do is to fork it and make the fixes we need to run React 16 then I'll make a PR for you guys 😄

PR #392, opened 2017-06-01
Fix fieldtooltip hover

kgilles:
Makes sure that the tooltip bubble stays visible when leaving the icon and entering the bubble itself.

### DIFF
--- a/Field/example.jsx
+++ b/Field/example.jsx
@@ -267,7 +267,7 @@ import Fieldset from '@klarna/ui/Fieldset'`,
           <Field left square size='1/3' label='Your favorite pokémon' />
           <Field right square size='2/3' label='Android or iPhone: please explain' />
           <Field left bottom size='2/5' label='Date of Birth' />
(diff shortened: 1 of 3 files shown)


xaviervia:
Thanks for the example. Unfortunately the fix doesn’t seem to be working though.
tooltip-dissappears

kgilles:
You have to go via the arrow. That's the best I could do with the Tooltip component as I couldn't figure out how to expand the Tooltip bubble surface closer to the icon without putting the entire thing too close to the icon itself.

xaviervia:
One idea is to add a transparent `div` covering the area just below the bubble as part of the tooltip, something like:

artboard

That way the target area would be bigger and it would still be active if the mouse is in the vicinity.

What do you think?

kgilles:
I'll see if I can make that happen.

kgilles:
I added a transparent span below the bubble, but didn't make it 100% width because I felt like it shouldn't interfere too much with the space of the actual field it's inside.
screen shot 2017-06-05 at 11 38 04 am

PR #397, opened 2017-06-20
Move preview link

beveryday:


xaviervia:
The content overflows from the top :) I would guess the `float` is affecting it's layout, since it doesn’t happen when there is no link.

screen shot 2017-06-20 at 10 34 51

PR #398, opened 2017-06-20
Add terms banner

beveryday:


TermsBanner/styles.scss
```diff
@@ -0,0 +1,25 @@
+@import '../css/settings';
+@import '../css/mixins/index';
+
+.banner {
+  background-color: map-get($colors, light-grey);
+  color: map-get($colors, grey-text);
+  line-height: 17px;
```
xaviervia: Both the text and the box are a little outside the grid :)

screen shot 2017-06-21 at 15 14 43

Fieldset/styles.scss
```diff
@@ -2,7 +2,9 @@
 @import '../css/mixins/index';
 
 .fieldset {
+  background-color: white;
```
beveryday: The problem is when how they want this to render. If you look at the design they want the gray to go above the corners of the address panel.

beveryday: screen shot 2017-06-26 at 9 34 01 am

beveryday: screen shot 2017-06-26 at 10 41 55 am

beveryday: If there is another way to fix this I will gladly use that instead.

Deschtex: Yea, we should avoid adding fixed background colors for things like this. If the border radius is changed through customization, this would not work anyway.

PR #402, opened 2017-06-27
Update @klarna/higher-order-components

xaviervia:
This might even work


cskeppstedt:
feelsgood

Such consistent naming. 👍  :shipit:

PR #404, opened 2017-06-29
Add prefilled prop

beveryday:


beveryday:
Add examples, so easy, so clean

PR #405, opened 2017-06-30
Allowing terms banner to accept style prop

beveryday:


beveryday:
Allowing terms banner to accept style prop

PR #406, opened 2017-07-03
Update overridable

xaviervia:


Deschtex:
LGTM, but perhaps it's wise to have someone else review it as well

PR #417, opened 2017-08-03
Fix inline icons so they obey custom styling

WebCloud:
Add `style` parameter so the icons can have custom `fill` and `stroke` with a custom theme/styling.


Field/styles.scss
```diff
@@ -326,6 +326,7 @@
 
   &__fill {
     @include illustration__fill(map-get($colors, grey-icon));
+    fill: 000;
```
xaviervia: It's this missing a `#` ?

WebCloud: that should not be there. It's a test that I've forgot to revert. It's now fixed on d63c97a9c322cbdaef0a9ddf43d3f983c78849cc

PR #419, opened 2017-08-09
Add more customization options to the Field component

WebCloud:
Add option to customize the `border-shadow` when focused and also enhance the inline icon colouring.

### DIFF
--- a/Field/index.jsx
+++ b/Field/index.jsx
@@ -225,7 +225,7 @@ const Field = React.createClass({
           borderColor: this.state.hover || focus
             ? customize.borderColorSelected
             : customize.borderColor,
-          boxShadow: focus && `0 0 4px ${customize.borderColorSelected}`
+          boxShadow: focus && (customize.boxShadow ? customize.boxShadow : `0 0 4px ${customize.borderColorSelected}`)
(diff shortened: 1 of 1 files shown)


Field/index.jsx
```diff
@@ -225,7 +225,7 @@ const Field = React.createClass({
           borderColor: this.state.hover || focus
             ? customize.borderColorSelected
             : customize.borderColor,
-          boxShadow: focus && `0 0 4px ${customize.borderColorSelected}`
+          boxShadow: focus && (customize.boxShadow ? customize.boxShadow : `0 0 4px ${customize.borderColorSelected}`)
```
xaviervia: Where is `customize.boxShadow` coming from? In the `withTheme` HoC is not being set

WebCloud: I've added directly there. I guess I should add to the HoC instead...

xaviervia: Probably both :D

WebCloud: Just for future reference, as discussed in person, these changes are ok to be carried as this for those customizations won't be officially supported by the Theme/Merchant Theming features.

PR #436, opened 2017-10-02
Fix Dialog not being centered

WebCloud:


WebCloud:
This will introduce some drawbacks with the vertical alignment. So I'll come up with a fix on the host app.

PR #437, opened 2017-10-02
Enable React 16: Replace React.createClass with create-react-class

stipsan:
This enables React 16 support as far as I've seen.


stipsan:
@xaviervia you think you're able to give this a look any time soon? 😄

schold:
@stipsan Nice, thanks. Would you mind also updating the `import` React, {PropTypes} from 'react'` in `Strong/index.jsx` to use prop-types?

schold:
Thanks, I'll have a look.

icons/index.jsx
```diff
@@ -5,7 +5,7 @@ export { default as Cancel } from './Cancel'
 export { default as Calendar } from './Calendar'
 export { default as Card } from './Card'
 export { default as Check } from './Check'
-export { default as Chevron } from './Chevron'
+export * as Chevron from './Chevron'
```
skogby: We need to be ES6 compatible, which this is not... could it be reverted or 
```
import * as chevron from './Chevron';
export { chevron };
```

stipsan: Hey @skogby.
If the intention is to allow this:
```jsx
import {Chevron} from '@klarna/ui'

export default () => <Chevron.Down />
```

If we revert to the old behavior then the above code breaks. The `Down` component in `./Chevron` is a `named export`. It's not a property on the default export. This code might work on older versions of webpack and babel, but upgrade to webpack v4 and babel v7 and you'll see that it breaks as it's not standard.

I have enabled "Allow edits from maintainers" so you should be able to push changes to my PR, let me know if you don't. My company have made some tech stack decisions (like moving everything to react 16.3 and use styled-components everywhere) so we are transitioning away from klarna ui. I likely won't have time to continue working on this PR so you folks will have to decide wether you want to close this PR or continue where I left off 😄

skogby: I understand! I can merge this PR and make the changes needed to make it run in bable 3 as well.

